PR #2, opened 2015-05-06
Initial Release allowing blink sketch compile and download

climberhunt:


bridiver:
@climberhunt https://github.com/arduino/ArduinoCore-arc32/commit/09a6c2ea91c99a7f8e3aff45cb4bc7f3e501a715#r160720685

image

I don't see any tests and now there is also probably code that has been adapted for this non-standard behavior so maybe  a flag could be added to disable this behavior?

PR #61, opened 2015-10-29
Fix BLE library String handling

facchinm:
The correct way to get a String lenght is by calling String.length()
Now len field is protected so a compilation error will be triggered if accessing the field directly


Dan-Emutex:
Looks good to me, thanks Martino.

@calvinatintel Ready for merge

PR #64, opened 2015-11-06
Resolving KW issue with SoftwareSerial destructor

bbaltz505:
Class has virtual functions inherited from a base class, but its destructor is not virtual and not empty


bbaltz505:
@bigdinotech please review

bigdinotech:
@bbaltz505 @calvinatintel looks good to me.

calvinatintel:
Merged in by mistake. I've reverted master to the previous commit but it doesn't seem like the PR reopens automatically. I'll merge in again.

calvinatintel:
Merged

PR #70, opened 2015-11-25
Add error message if using IDE 1.6.6

facchinm:
`arduino-builder` creates buggy binaries on 1.6.6 only, so check if we are using that version and break the compilation.
Nightly versions are already tagged 1.6.7 so they are safe to use (also 1.6.5 is safe, too)

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -29,6 +29,10 @@
 #include "binary.h"
 //#include "itoa.h"
 
+#if ARDUINO == 10606
+#error IDE version incompatible with Arduino 101, please upgrade to it to version 1.6.7 or newer
+#endif
+
 #ifdef __cplusplus
 extern "C"{
 #endif // __cplusplus


calvinatintel:
@bigdinotech @kmsywula @intel-larry @SidLeung Please review

SidLeung:
@bigdinotech  @kmsywula  @intel-larry Shouldn't be <= 10606 instead of == 10606?

calvinatintel:
The problem doesn't exist on 1.6.5 or 1.6.4. I think 10606 is correct.

calvinatintel:
Well actually our other merges break versions prior to 1.6.6, so <= would be correct.

bigdinotech:
I think it should be <=10606 since we are merging in the new recipe pattern

calvinatintel:
Since Martino is probably asleep, I'll make a new PR under my fork with <=

calvinatintel:
Replaced with https://github.com/01org/corelibs-arduino101/pull/71

PR #73, opened 2015-11-26
CurieRTC - SetTime example not setting RTC

bigdinotech:
RTC not being set because the wrong call was being made should call
RTC.set(t) to set the hardware RTC instead of setTime(t).


bigdinotech:
@kmsywula @SidLeung @intel-larry pls review
@calvinatintel pls merge once ready.

kmsywula:
@bigdinotech syntax looks correct for me; I cannot really comment more on that, we need to check if it compiles fine

calvinatintel:
cherry-picked

PR #76, opened 2015-12-02
modified example in Arduino style with accompanying Processing sketch

hbisby:
The RawIMUSerialData example has been modified to the code of a tutorial written which uses the IMU data to calculate Euler angles and present the orientation of the board in Processing (sketch is commented at the bottom of RawImuSerialData.ino). Ifdefs have been removed (replaced by if statement for calibration) as well as the option for binary output, since the format was considered too complicated for Arduino style. Types have also been changed to Arduino, e.g `uint16_t` has been changed to `int`.

### DIFF
--- a/libraries/CurieImu/examples/RawImuDataSerial/RawImuDataSerial.ino
+++ b/libraries/CurieImu/examples/RawImuDataSerial/RawImuDataSerial.ino
@@ -1,147 +1,232 @@
 /*
-===============================================
-Example sketch for CurieImu library for Intel(R) Curie(TM) devices.
-Copyright (c) 2015 Intel Corporation.  All rights reserved.
+  ===============================================
+  Example sketch for CurieImu library for Intel(R) Curie(TM) devices.
+  Copyright (c) 2015 Intel Corporation.  All rights reserved.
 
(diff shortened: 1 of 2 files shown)


sandeepmistry:
@facchinm @tigoe do we want this change to be included? If so, I can add rebasing it and updating to my list of things to look at.

tigoe:
I'd rather see the example updated to the new IMU API first.

kitsunami:
@tigoe, is this pull request still valid?

tigoe:
Probably, but I’d defer to Sandeep on it. Helena made these changes first, then Sandeep made bigger changes. I assume he had incorporated all of her changes, but he could say for sure. On the road at the moment, and not able to look at the full diff. @sandeepmistry, comments?
Tom

sandeepmistry:
@kitsunami please close this, instead of bundling an example that depends on a the external  MadgwickAHRS library, we will make this into a tutorial.

@tigoe the MadgwickAHRS library is used to calculate the yaw, roll and pitch from the raw accelerometer and gyroscope values. So, it won't work out of the box unless the user adds the library on their system.

cc/ @facchinm @agdl

PR #81, opened 2015-12-04
Initial port of BLEPeripheral style API to CurieBle library

sandeepmistry:
Initial port of the [BLEPeripheral](https://github.com/sandeepmistry/arduino-BLEPeripheral) API to the CurieBle library.

Changes also include:
* Internal updates to allow for no limit on the number of services, characteristics, and descriptors that can be added to the peripheral.
* The buffers used for characteristic and descriptor values are not dynamically allocated.
* keywords.txt file added for syntax highlighting support in the Arduino IDE.

Example updates:
 * BatteryMonitor updated to Arduino example style
 *  Added 3 new LED based examples (LED, LED callback, LED with button)
 * AutomationIO example has been removed as it was to complex of an example for the targeted user base. The new LED examples cover how to achieve the same goal with a single pin.

Notes:
 * There is no services changed characteristic, so iOS will cache previous services and characteristics when they change in a sketch upload. Turning Bluetooth off and then back on again in iOS clears the cache. I believe a fix for this is needed on the firmware side. (This was already an issue with the previous version of the library).
 * Some of the BLEPeripheral API's have not been ported over as they will require support on the nRF firmware. They can be added in a later release.

cc/ @tigoe @hbisby @facchinm


bigdinotech:
@sandeepmistry looks great thanks!
@calvinatintel please merge

PR #86, opened 2015-12-15
Jira-32, Jira-386

SidLeung:
A new library, CurieArcTimerOne, in place of the TimerOne library originated from, and applicable to, the AVR platform.  The library provides both timer and software PWM utilities.


calvinatintel:
@kmsywula @bigdinotech Please review

bigdinotech:
Yeah maybe CurieTimerOne would be a better name?
Lets ask Arduino for their input.
@tigoe what do you think?

SidLeung:
The reason I gave the library a strange name, CurieArcTimerOne, was due to the fact that there was a timer-1 in Curie (actually 4 of them) that can also be used for PWM and there was a timer-1 in the ARC subsystem of Curie.

tigoe:
Sorry, on planes and can't see the whole discussion on phone. CurieTimerOne
sounds better at first glance, but I'd defer to @mellis or @sandeepmistry
or someone who can read the whole thread before me.

sent on the go. please excuse brevity and mistaken auto corrections

SidLeung:
@bigdinotech @kmsywula Modification made according to Dino's input.  Please review.  Waiting on name decision.

sandeepmistry:
I would suggest using the name ```CurieTimer```.

The ```One``` and ```Three``` in Paul's [```TimerOne```](https://github.com/PaulStoffregen/TimerOne) and [```TimerThreee```](https://github.com/PaulStoffregen/TimerThree) library seems to map to the AVR timer register used and only support a single timer each. Where as this new library has support for 2 timers (0 and 1).

SidLeung:
Please note that the Arc timer-0 of the Curie is used for timing delays already and, thus, a static object (singleton) of the Arc timer-1 is created.  In addition, the Curie SoC has 4 other timers that have PWM capability.  Tried to use the name of the library to highlight these two pieces of info.  May be that is not desirable,

sandeepmistry:
Good to know, I suggest we remove the ability for the users to create an instance using ```ARC_TIMER_0```, as well as remove the constant from the header. If timer-1 will only be supported, then the name ```CurieTimerOne``` suggested by @bigdinotech works.

SidLeung:
@bigdinotech @kmsywula Library name changed.  Please review.

sandeepmistry:
@SidLeung there are still a few things that need to be renamed:
 * the name value in ```library.properties```
 * filenames are still ```CurieArcTimer.cpp/h```
 * instance and class names still contain "Arc"

SidLeung:
@bigdinotech @kmsywula Requests from Sandeep made.  Please review.

SidLeung:
@bigdinotech @kmsywula Corrected naming in keywords.txt and added method names.

SidLeung:
@bigdinotech @kmsywula Please review.

bigdinotech:
@calvinatintel please merge or cherry-pick

calvinatintel:
Cherry-picked

PR #88, opened 2015-12-17
hardcode sketchUploader version

facchinm:
while we solve the bug IDE-side, it is safe to revert using the hardcoded version number to handle the sketchUploader installation;
This solves a clash with outdeted Galileo/Edison core which tools are picked randomly (depending on json download order and java filesystem library internal ordering).

@calvinatintel

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -89,16 +89,16 @@ recipe.size.regex=Total\s+([0-9]+).*
 
 # Arc Uploader/Programmers tools
 # -------------------
-tools.arduino101load.cmd.path={runtime.tools.sketchUploader.path}/clupload/cluploadArduino101_osx.sh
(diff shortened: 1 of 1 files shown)


calvinatintel:
@kmsywula @bbaltz505 @bigdinotech @SidLeung @intel-larry @novashah Please see the urgent message on basecamp

bigdinotech:
@calvinatintel please merge

bbaltz505:
So this file will update on every release? Will we do it manually or let Jenkins build process search/replace the string?

bigdinotech:
I think the plan is to this for now while the bug is fixed in the IDE side.
We can also formulate a cleaner way to this on all of the boards while that happens

calvinatintel:
Until the IDE is fixed, this file will update each time the tools change.

PR #101, opened 2016-01-08
CurieSoftwareSerial Improvements

bigdinotech:
-renamed global variable rxPin to _rxPin
-allow different baud rates for different instances of a SoftwareSerial object
-minor code cleanup of unused code

### DIFF
--- a/libraries/CurieSoftwareSerial/src/SoftwareSerial.cpp
+++ b/libraries/CurieSoftwareSerial/src/SoftwareSerial.cpp
@@ -42,21 +42,17 @@ char SoftwareSerial::_receive_buffer[_SS_MAX_RX_BUFF];
 volatile uint8_t SoftwareSerial::_receive_buffer_tail = 0;
 volatile uint8_t SoftwareSerial::_receive_buffer_head = 0;
 
-
-//
-// Globals
-//
-uint8_t rxPin;
-uint16_t bitDelay;
-uint16_t rxIntraBitDelay;
-int rxCenteringDelay;
-int initRxIntraBitDelay;
-int firstIntraBitDelay;
-int initRxCenteringDelay;
-bool firstStartBit = true;
-bool bufferOverflow = true;
-bool invertedLogic = false;
-bool isSOCGpio = false;
+static uint8_t _rxPin;
+static uint16_t bitDelay;
+static uint16_t rxIntraBitDelay;
+static int rxCenteringDelay;
+static int initRxIntraBitDelay;
+static int firstIntraBitDelay;
+static int initRxCenteringDelay;
+static bool firstStartBit = true;
+static bool bufferOverflow = true;
+static bool invertedLogic = false;
+static bool isSOCGpio = false;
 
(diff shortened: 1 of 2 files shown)


bigdinotech:
@kmsywula @SidLeung ready for code review

libraries/CurieSoftwareSerial/SoftwareSerial.cpp
```diff
@@ -94,14 +94,21 @@ bool SoftwareSerial::listen()
     bufferOverflow = false;
     _receive_buffer_head = _receive_buffer_tail = 0;
     active_object = this;
-    rxPin = _receivePin;
+    _rxPin = _receivePin;
+    rxIntraBitDelay = _rx_delay_intrabit;
+    rxCenteringDelay = _rx_delay_centering;
```
soundanalogous: Is there a good reason for defining any global variables here? This essentially makes all of these variables "reserved words" for anyone who includes this library in their sketch.

bigdinotech: The reason I used globals is because recv() which acts as the interrupt handler is static.
Therefore, it cannot access those variables if I kept them as member variables of the class.

Do you think having those variables names "reserved" can be an issue? If so I can always rename them to something else.

Also if you have a better solution, Pull Requests are always welcome. :)

kmsywula: If they are internal and only to be used inside the library, the convention is to precede name with _ character

Also if you made them static then scope is limited to the file they are defined in.

PaulStoffregen: just make them static.

SidLeung: One other way, equally not pretty, to allow static ISR routine to access variable of an object is,
1.  Define the ISR as private in a Class.
2.  Instantiate a static object of the Class.
3.  Define a static function, ISR_wrapper, that the sole purpose is to call the ISR of the Class via the static object.
The only advantage here is that the ISR is now free to access all variables in the Class without making them global, static, etc.

soundanalogous:
Using an unnamed namespace may also work (but I haven't tried this within an ISR that could be a limitation). Wrap all of the global variables with:

```c++
namespace {
  // declare your variables here
}
```
The compiler will create a unique namespace. There is nothing else you'd need to change. Here's an article that explains this approach: http://www.comeaucomputing.com/techtalk/#nostatic

SidLeung:
Mmm...  May be not.  ISR is not the direct cause.  The fact that the ISR has to be declared as a static routine and it needs to access data/variables in a class that has not been instantiated, the compiler get upset.  Static function can only access static variable...

bigdinotech:
@kmsywula @SidLeung please review again

SidLeung:
@calvinatintel Looks good.  All global are static.

calvinatintel:
Cherry-picked

PR #102, opened 2016-01-12
[DO NOT MERGE YET] Use unified upload tool

facchinm:
By replacing existing `bash` + script procedure (very OS-specific and error prone) with this tool (https://github.com/facchinm/arduino101load) the flasher commandline generation can be greatly simplified.
This would also help separation between legacy `sketchUploader` and dfu-based uploader for 101

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -89,16 +89,9 @@ recipe.size.regex=Total\s+([0-9]+).*
 
 # Arc Uploader/Programmers tools
 # -------------------
-tools.arduino101load.cmd.path={runtime.tools.sketchUploader-1.6.4+1.15.path}/clupload/cluploadArduino101_osx.sh
-tools.arduino101load.cmd.path.linux="{runtime.tools.sketchUploader-1.6.4+1.15.path}/clupload/cluploadArduino101_linux.sh"
-tools.arduino101load.cmd.path.windows={runtime.tools.sketchUploader-1.6.4+1.15.path}/x86/bin/bash.exe
-tools.arduino101load.cmd.script.windows={runtime.tools.sketchUploader-1.6.4+1.15.path}/clupload/cluploadArduino101_win.sh
-tools.arduino101load.cmd.bin.windows={runtime.tools.sketchUploader-1.6.4+1.15.path}/x86/bin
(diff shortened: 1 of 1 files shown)


bbaltz505:
Thanks. I'll try it out. So we are only distributing a OS-specific Go binary for the script then? That is much simpler. I suppose a downside is that you can't view and tweak the upload script if you had to.

facchinm:
Right, but the source is very straightforward. 
I also forked the tools repo here (https://github.com/facchinm/intel-arduino101-tools), removing all the edison/galileo related files and retagging to work in the Board manager, if you wish to test it we could merge before the next core release :smile:

calvinatintel:
I like it, but we can't remove the support files like `bash.exe` because Edison and Galileo use them.

calvinatintel:
@facchinm, one of the downside of this approach is that the uploader becomes less user-serviceable compared to a `.sh`. Do you foresee a problem with this?

SidLeung:
I would like to know the original reasoning/purpose of choosing scripting over an utility program/app to invoke the tools.  Is it,
  +  Visibility is important to the end user.
  +  Max flexibility for the end user.
  +  DIY changes for the advance user.
  +  The cheapest and fastest way to get the system going.

If it is the last reasoning, which is very valid, it would be a partial solution.

facchinm:
Thanks for the comments, with this PR and the twin tools repo I tried to resolve a couple of issues
* bash scripts is an hell when handling file paths and complex operations (the original script was broken for paths with spaces and needed a temp file to parse the `dfu` command output)
* there is no need to have different paths for different OS
* Edison/Galileo uploader is 100% different from 101 loader, I see no good reason to have them contained in a single package
The "hackability" of a script is higher, of course, but since the replacement is open source and setting up the go toolchain is a 10 minutes task, it could reach the same objective with a much cleaner implementation

calvinatintel:
@facchinm We like the golang solution, but it might be more suitable for next product. Do you think the current implementation should be replaced by a golang version?

bbaltz505:
Closing this PR as we will not use this tool for Arduino 101. (see PR 104)

PR #103, opened 2016-01-15
Remove OneWire and Adafruit_DotStar libraries.

pricopb:
Code changes necessary to port them to Arduino101 moved to libraries repositories, therefore they should be removed from corelibs, not being maintained  here.


bigdinotech:
I see the changes committed to OneWire and Adafruit Github repos.
@calvinatintel please merge

PR #104, opened 2016-01-19
ATLEDGE-491 - rename sketchUploader tool to arduino101load

bbaltz505:
Renaming sketchUploader tool to arduino101load to avoid conflict in IDE if Galileo/Edison is installed

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -89,16 +89,16 @@ recipe.size.regex=Total\s+([0-9]+).*
 
 # Arc Uploader/Programmers tools
 # -------------------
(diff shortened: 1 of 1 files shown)


bbaltz505:
@kmsywula @calvinatintel @SidLeung please review. You can try test build:
http://mkfs.ndg.intel.com/precommit/json/package_0.0.14_index.json

calvinatintel:
It works on OSX, but it may also be because maintainer is `Intel-Test` which is enough to avoid the conflict.

PR #109, opened 2016-01-26
Jira-501: CurieTimer.resume() did not re-enable interrupt at the controller.

SidLeung:
CureTimerOne library:  Routine resume() did not re-enable interrupt upon exit - bug.  Corrected.


SidLeung:
@bigdinotech @kmsywula Please review.

PR #117, opened 2016-02-05
Jira-507 and 514: 1. Bug corrected in timer ISR where the clearing of…

SidLeung:
… the pending interrupt was at the entrance instead of exit.  When interrupts could happen fast enough to put the ISR in a loop. 2. The overhead for the s/w pwm code is 4-5 usec.  Thus, limited the shortest pwm pulse to 10 usec (high and low).


SidLeung:
@bigdinotech Please review.

calvinatintel:
Cherry-picked

PR #118, opened 2016-02-10
Stop interrupt when switching to 100% duty cycle

bbaltz505:
This fix is similar to the 0% duty-cycle fix. The interrupt was not being stopped for 100% duty-cycle condition.


SidLeung:
@calvinatintel Looks good.  Please merge.

PR #122, opened 2016-02-23
use sprintf-based implementation for dtostrf

facchinm:
altough the current implementation passed all the unit tests, an user reported problems with the actual code (http://forum.arduino.cc/index.php?topic=380422.0).
If libc has float support there is no need to do strange things like on AVR, so we can use the sprintf implementation


kitsunami:
@bigdinotech please review.

bigdinotech:
@calvinatintel looks good. please merge

calvinatintel:
cherry-picked

PR #125, opened 2016-02-25
Remove custom Adafruit_Motor_Shield_V2 library

bbaltz505:
Now that 1.0.4 version of the public library supports the 101, we can remove this from our corelibs


bbaltz505:
@calvinatintel please review

PR #126, opened 2016-02-25
Fixing sketch size calculation (text + ctors + rodata)

bbaltz505:
Change the total sketch size calculation based on Sidney's explanation:

> The arc-elf32-size.exe added in other sections to the total Sketch flash usage and that is not correct.  The calculation should be the sum of text, ctor, and rodata sections.  Need to correct the upload script to reflect that.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -11,6 +11,7 @@ arduino_101.vid.0x2A03.warning=Uncertified
 arduino_101.upload.tool=arduino101load
 arduino_101.upload.protocol=script
 arduino_101.upload.maximum_size=196608
+arduino_101.upload.maximum_data_size=8192
 arduino_101.upload.use_1200bps_touch=true
 arduino_101.upload.wait_for_upload_port=false
 arduino_101.upload.native_usb=false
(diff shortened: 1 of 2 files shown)


bbaltz505:
Size breakdown using arc-elf32-size:
<pre>
section     size         addr
text       13136   1073954816
ctors         12   1073967952
rodata       440   1073967964
datas       3156   2818629632
bss         1916   2818632788
heap        8192   2818634704
stack       2560   2818642896
.comment     118            0
Total      29530

Sketch uses 13,588 bytes (6%) of program storage space. Maximum is 196,608 bytes.
</pre>

bbaltz505:
According to Sidney Leung:

> As for providing dynamic RAM usage, it is not a practical exercise. First of all, the Arduino platform does not have an OS on the ARC side. Thus, dynamic memory allocation is simply finding a large enough piece of memory from the Heap to use - that's done by malloc(). There is minimal amount of de-fragmentation processing in malloc - best it can do is to join two pieces of adjacent free-ed memory together into a larger piece. Therefore, the sum of unused/free-ed RAM is not the amount of memory available for malloc. This renders the efforts of returning available memory to user meaningless.

PR #119, opened 2016-02-17
CurieIMU API changes

sandeepmistry:
* Remove enums for ranges and duration, use floats and map to closest BMI160 enum
* Set and get threshold API's now accept float and map to threshold based on accelerometer range
* Auto calibrate and set offset now enables offset
* Change interrupt routines to ```interrupts``` and ```noInterrupts```
* keywords.txt updates

PR #128, opened 2016-02-26
Change set/get accelerometer and gyro offset API's to use floats values

sandeepmistry:
This makes it consistent with other API changes in #119 to convert the register values to floats (mg, degrees/second) and vice versa.


bigdinotech:

calvinatintel:
cherry-picked

PR #129, opened 2016-03-02
Moved .h and .cpp file in src folder

00alis:
CurieTimerOne, SPI, Servo, and Wire libraries were missing the 'src' folder. 
This causes conflicts in the way we display these libraries in the Arduino Create Editor :)
Added folder.

@bigdinotech


matteosuppo:
Actually another issue is that the library.properties file is called libraries.properties

00alis:
Updated that too!

bigdinotech:
@calvinatintel looks good. Please merge.

PR #131, opened 2016-03-03
README.md: Add note to specify that no other folders can be present i…

### LINKED ISSUE #130: Error when using latest version from git on linux following the instructions
I put this in the place of the 1.0.4/ board folder and when I hit 'verify' I get this:

```Board arduino_101 (platform arc32, package Intel) is unknown

Error compiling.```

Not really too sure what I'm doing wrong :/

arfoll:
…n arc32

Having another folder causes a compilation error with an unknown board error.

Fixes #130

Signed-off-by: Brendan Le Foll <brendan.le.foll@intel.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,6 @@ If you wish to use the latest **untested** changes, follow these instructions.
   * OS X: `~/Library/Arduino15`
   * Linux: `~/.arduino15`
 5. Go to `packages/Intel/hardware/arc32/<version>/`
-6. Delete the content of the directory from step 5, and replace it with the content of the "corelibs-arduino101-master" folder in the zip from step 2.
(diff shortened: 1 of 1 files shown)


calvinatintel:
`Arduino15` directory is not expected to be manually edited. Adding extra directory is one of many ways to break the system. As such, the users are expected to follow the direction    
6. Delete the content of the directory from step 5, and replace it with the content of the "corelibs-arduino101-master" folder in the zip from step 2.    
I'm not sure if more needs to be said.

arfoll:
Your choice :) it confused me so I assumed others may be too.

But if there never is supposed to be another folder one wonders why a folder is there in the first place...

calvinatintel:
I understand, and I'll keep it in mind. Let's see how the community feels.

PR #153, opened 2016-03-27
Adding settling time to analogRead().

ljardo:
The original analogRead() function switches the mux channel and initiates the ADC conversion immediately. However, this causes settling errors when the source impedance is in the kohm range. Insert at least 10 us of delay between switching the channel and initiating the conversion to eliminate the settling error.

### DIFF
--- a/cores/arduino/wiring_analog.c
+++ b/cores/arduino/wiring_analog.c
@@ -130,8 +130,13 @@ uint32_t analogRead(uint32_t pin)
     SET_ARC_MASK(ADC_CTRL, ADC_SEQ_PTR_RST);
     /* Update sequence table */
     WRITE_ARC_REG(p->ulAdcChan, ADC_SEQ);
-    /* Reset sequence pointer & start sequencer */
-    SET_ARC_MASK(ADC_CTRL, ADC_SEQ_PTR_RST | ADC_SEQ_START | ADC_ENABLE);
+    /* Reset sequence pointer & enable the ADC */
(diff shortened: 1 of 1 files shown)


cores/arduino/wiring_analog.c
```diff
@@ -130,8 +130,13 @@ uint32_t analogRead(uint32_t pin)
     SET_ARC_MASK(ADC_CTRL, ADC_SEQ_PTR_RST);
     /* Update sequence table */
     WRITE_ARC_REG(p->ulAdcChan, ADC_SEQ);
-    /* Reset sequence pointer & start sequencer */
-    SET_ARC_MASK(ADC_CTRL, ADC_SEQ_PTR_RST | ADC_SEQ_START | ADC_ENABLE);
+    /* Reset sequence pointer & enable the ADC */
+    SET_ARC_MASK(ADC_CTRL, ADC_SEQ_PTR_RST | ADC_ENABLE);
+    /* Insert some delay to allow the channel to fully settle */
+    /* Here, the delay is 1 ms, but 10 us should suffice for <10 kohm source impedance */
+    delay(1);
```
eriknyquist: So this is adding 1ms of overhead each time analogRead is called, even if the source impedance is low enough to not require it? That doesn't sound great to me. Can we figure out a way to make this an option? i.e. if the user knows they will have impedance in the K'ohms, they can enable the settling delay

eriknyquist:
@ljardo can you share some more details about this issue please?

specifically, 1) the sampling rate used and 2) the code used to test

eriknyquist:
OK, I see no good reason for this fix and there's been no activity by @ljardo since the original posting so I'm going to close this.

PR #154, opened 2016-03-31
Jira-575, Jira-506:  Bug fixes to print float and double correctly, i…

SidLeung:
…n a string or std output.

With latest toolChain, the printing of float and double would result in an overfloat condition if the value is bigger than 8 digits.  The String construction would hang the code.  The fix put it will print out a large float/double in the exponent format of, x.xxe+yyy, if the value has more than 8 digits.

@calvinatintel @bigdinotech @kmsywula   Please review my changes.

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -289,18 +289,29 @@ size_t Print::printLongLong(unsigned long long n, uint8_t base) {
 size_t Print::printFloat(double number, uint8_t digits)
 {
   size_t n = 0;
+  int exponent = 0;
+  double tmp;
 
   if (isnan(number)) return print("nan");
   if (isinf(number)) return print("inf");
-  if (number > 4294967040.0) return print ("ovf");  // constant determined empirically
-  if (number <-4294967040.0) return print ("ovf");  // constant determined empirically
 
   // Handle negative numbers
-  if (number < 0.0)
-  {
-     n += print('-');
-     number = -number;
+  if (number < 0.0) {
+    n += print('-');
+    number = -number;
   }
+
+  // Chk if integer part has more than 8 digits.
+  tmp = number;
+  while (true) {
+    tmp /= 10.0;
+    exponent++;
+    if (tmp < 10.0)  break;
+  }
+  if (exponent > 8)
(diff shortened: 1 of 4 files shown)


SidLeung:
@bigdinotech Please check out my changes.  The Flex team wants this code change so that they can check out the compiler/toolchain changes.  Thanks!

bigdinotech:
@calvinatintel code looks good to me. Please merge

SidLeung:
@bigdinotech Thanks, Dino.
@calvinatintel This are the bug fixes that are needed to release with the Floating Point Toolchain changes.  These changes work with the old toolchain, it's backward compatible.

facchinm:
Hi @SidLeung ,
I'm sorry but this patch in fact reverts #122 and  breaks the following sketch
```
float latitude = 45.20202;
float longitude = 8.656595;
char latBuffer[12];
char longBuffer[12];
char link[60];

void setup() {
  Serial.begin(115200);

  dtostrf(latitude, 0, 8, latBuffer);
  dtostrf(longitude, 0, 8, longBuffer);
  sprintf_P(link, PSTR("www.google.it/maps/place/%s,%s"), latBuffer, longBuffer);
}

void loop() {
  Serial.println(link);
  delay(100);
}
```
Maybe the new toolchain is needed for this to work correctly?

SidLeung:
I will check out the appended Sketch.  Thanks.

SidLeung:
Would you please provide the expected result of the Sketch?  What is the expected output if, say, latitude has a value of 2^32 = 4294967296?  Thanks.

facchinm:
Hi @SidLeung ,
the expercted output of the sketch is 
```
www.google.it/maps/place/45.20201900,8.65659520
```

Using  2^32 = 4294967296 as latitude on AVR we get
```
www.google.it/maps/place/4294967300.00000000,8.65659520
```

SidLeung:
Thanks for the quick answer.  However, as stated in the sketch, latBuffer is 12 char long. For, a string of, "4294967300.00000000", will overflow the buffer when it is used by dtostrf().  Can, sprintf_P, still construct the entire string so nicely?

PR #156, opened 2016-04-01
CurieI2S library

bigdinotech:
-interrupt based i2s library

### DIFF
--- a/libraries/CurieI2S/src/CurieI2S.cpp
+++ b/libraries/CurieI2S/src/CurieI2S.cpp
@@ -1,0 +1,363 @@
+//CurieI2S.cpp
+
+#include "CurieI2S.h"
+#include <interrupt.h>
+
+Curie_I2S CurieI2S;
+
+static uint32_t _i2s_dma_Rx_Buffer[DMA_BUFFER_SIZE];
+static uint32_t _i2s_dma_Tx_Buffer[DMA_BUFFER_SIZE];
+
+static void i2sInterruptHandler(void)
+{
+    //TODO: Make interrupt handling more atomic
+    noInterrupts();
+
+    //tx FIFO full
+    if(*I2S_STAT & 0x00000400)
+    {
+        //Serial.println("tx full");
+        
+        //disable TFIFO_FULL interrupts
+        *I2S_CID_CTRL = *I2S_CID_CTRL & 0xFBFFFFFF;
+        
+        //start transmit data in FIFO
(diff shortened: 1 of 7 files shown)


bigdinotech:
@SidLeung @kmsywula ready for code review

PaulStoffregen:
In your I2S_RxCallback.ino example, might be a good idea to check if count is greater than 255, to prevent a possible buffer overflow.

Even if the specific timing of this example code doesn't ever overflow the buffer, as an example published for general consumption by makers in the Arduino world, you can expect it to be used together with other example code that may alter the interrupt to main program timing.

bbaltz505:
Good catch, Paul. I forgot to Run the PR through Klocwork. It flagged possible buffer overflow on lines 137 and 324.

libraries/CurieI2S/src/CurieI2S.cpp
```diff
@@ -1,125 +1,270 @@
...
+                _i2s_Rx_BufferPtr->data[_i2s_Rx_BufferPtr->head] = data;
+                _i2s_Rx_BufferPtr->head = index;
+            }
+            #ifdef I2S_DEBUG
+            digitalWrite(I2S_DEBUG_PIN, HIGH);
+            digitalWrite(I2S_DEBUG_PIN, LOW);
+            #endif
+
+        }
+
```
SidLeung: May need to consider a timing window here.  Consider a byte came in, after fifoDataLength was set with the h/w register earlier on.  The FIFO would not be emptied here.  With the "Almost Full" interrupt disabled and the "Empty" interrupt enabled, there may no more interrupt from here on.  Would that be a real case?

bigdinotech: I actually think that it should be fine as long as we enter the ISR quickly enough.
The filling of the buffer is much faster than the transmission rate supported by the I2S.
The only real issue I see is if the user callback routine inside the ISR takes too long.

SidLeung: May be you leave the "Almost Full" interrupt enabled since the flag is going to be cleared.  If the FIFO got filled up because the call back was taking its merry time, the AFull interrupt would not be missed.

SidLeung:
@calvinatintel  Went over the Almost Full interrupt with Dino.  Looks good.  Please merge.

calvinatintel:
Used non fast-forward merge since there are multiple commits

PR #160, opened 2016-04-06
JIRA-578 pwmStop() leaves pin HIGH

russmcinnis:
CurieTimerOne::pwmStop() was doing a digitalWrite(pwmPin, LOW) after the kill() so unless the pin happened to be 0 the pwmPin might be left in the HIGH state causing motors or other devices to run HIGH after pwmStop(). Just removed the pwmPin = 0 from kill() and it passes all our loopback tests.

### DIFF
--- a/libraries/CurieTimerOne/src/CurieTimerOne.cpp
+++ b/libraries/CurieTimerOne/src/CurieTimerOne.cpp
@@ -76,7 +76,7 @@ void CurieTimer::kill()
   tickCnt = 0;
   userCB = NULL;
   currState = IDLE;
-  pauseCntrl = pauseCount = pwmPin = dutyCycle = nonDutyCycle = periodInUsec = 0;
+  pauseCntrl = pauseCount = dutyCycle = nonDutyCycle = periodInUsec = 0;
 }
 
 


russmcinnis:
@SidLeung 
Please review.
Jira: ATLEDGE-578

SidLeung:
@russmcinnis @bbaltz505 Oh!  That's a bug.  I agree with the fix.  However, can I entertain you with another solution.  The reason to zero out the pwmPin in kill() is to reset all the variables in the Class.  Instead of leaving out the reset of pwmPin in kill(), would you consider changing the pwmStop() to this,
{ digitalWrite(pwmPin, LOW); return kill() }
It is not a necessary change since put the pwmPin to 0 or leaving as it are just as bad.  Should choose an invalid port number but not kill the code...

bbaltz505:
@SidLeung is there a chance the interrup could immediately set the pin to HIGH between the digitalWrite and the kill()? Should we just have kill set the pin to LOW- instead of in Stop?

SidLeung:
@bbaltz505 @russmcinnis  Woo-Oh!  You are quite correct.  Can't do that.  Russ's changes are good.  Please merge @calvinatintel .

PR #164, opened 2016-04-13
check pin range first thing in pinMode(), digitalWrite() and digitalRead()

russmcinnis:
Check that pin number is not out of range in digitalRead(), digitalWrite() and pinMode()


russmcinnis:
@bigdinotech 
please review
ATLEDGE-579

bigdinotech:

PR #165, opened 2016-04-14
Retain unstripped ELF file for debugging

bbaltz505:
In order to debug with JTAG, an unstripped ELF file is needed. This change creates two ELF files: stripped and unstripped

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -70,6 +70,10 @@ recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compil
 ## Combine gc-sections, archives, and objects
 recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} "-T{build.variant.path}/{build.ldscript}" "-Wl,-Map,{build.path}/{build.project_name}.map" {compiler.c.elf.extra_flags} -o "{build.path}/{build.project_name}.elf" "-L{build.path}" "-L{build.variant.path}" -Wl,--whole-archive "-l{build.variant_system_lib}" -Wl,--no-whole-archive -Wl,--start-group "-l{build.variant_system_lib}" -lnsim -lc -lm -lgcc {object_files} "{build.path}/{archive_file}"
 
+## Save output with debug symbols (.debug.elf file)
+recipe.hooks.objcopy.preobjcopy.1.pattern=cp -f "{build.path}/{build.project_name}.elf" "{build.path}/../arduino101_sketch.debug.elf"
+recipe.hooks.objcopy.preobjcopy.1.pattern.windows=cmd /C copy /y "{build.path}\{build.project_name}.elf" "{build.path}\..\arduino101_sketch.debug.elf"
+
 ## Create output (.bin file)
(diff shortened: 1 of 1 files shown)


bbaltz505:
@SidLeung @kmsywula please review. I verified that the ARC gdb can read the unstripped ELF file created by the IDE.

SidLeung:
@bbaltz505 @calvinatintel Is there any reason you can think of that the elf file should not be created every time a Sketch is compiled?  I vote to merge the change into the main trunk.

bbaltz505:
These files normally get deleted after flashing. In the Linux IDE, there is an option in preferences to keep the binaries. I don't think it hurts to always keep a debug binary around.

SidLeung:
@bbaltz505 Will that option in "Preferences" affects the elf file too?  If so, the user has a choice.  Let's merge it.

bbaltz505:
Windows seems to always keep the temporary build folder (e.g., C:\Users\babaltz\AppData\Local\Temp\build5a12ec7801f914e3a9c773f3624a5617.tmp/BarometricPressureSensor.ino.elf)
On the next sketch compile, that folder seems to get cleared out and used again.

SidLeung:
I think it is a new folder created and the old one deleted.  Not the same folder.  At least the number part of the folder is changed.  Am I correct?

bbaltz505:
I believe the folder location is created when the IDE starts up. If you close and reopen your browser, it will use a new temp folder.

bbaltz505:
adding @bigdinotech for review

bigdinotech:
looks good to me. @calvinatintel please merge

bbaltz505:
I wonder if I should save the debug ELF somewhere more accessible - like the sketch folder? Otherwise the user has to have verbose compilation turned on and then copy the lengthy ELF path to use with the debugger.

bigdinotech:
That's a good idea to place the elf file in the same folder where the user saved it.
Only thing is they can write a sketch with out saving it and in that case where would we save the elf file?

SidLeung:
Yeah!  Save it to the Sketch directory.  Like it.  If the user does save the Sketch...  Well, the elf doesn't mean much to the user anyway.

bbaltz505:
Sorry Sidney. Actually now I don't like the idea of ELF files being left around in the sketch folder. That could be confusing to beginning users. Only advanced users are going to be debugging so they'll know how to find the ELF (via compile messages in the console window).

SidLeung:
How about a sub-directory under the user sketch folder, say, gdb_info?

bbaltz505:
According to Martino, the IDE does not have a pointer to the sketch folder. For now, I think its best to save the file in the build folder.
I updated the commit to just to copy the unstripped ELF before it's stripped. The command should support cross-platform.

bbaltz505:
@facchinm here is the config to save a debug ELF

bbaltz505:
@SidLeung ready for review. This change saves the debug elf file as {build.path}/../arduino101_sketch.debug.elf
e.g., C:\users\babaltz\appdata\local\temp\arduino101_sketch.debug.elf

SidLeung:
@bbaltz505 @calvinatintel Looks good.  BTW, what the location for Ubuntu?  Please merge.

bbaltz505:
I believe temporary build folders are in /tmp so the debug ELF file would be:
/tmp/arduino101_sketch.debug.elf
These debug ELF files are not automatically cleaned up when the IDE closes and are overwritten on each sketch build.

bbaltz505:
rebased onto latest master

PR #167, opened 2016-04-14
Jira-584: String constructor create floating point string with incorr…

SidLeung:
…ect number of decimal places.  The "." was counted as one decimal place => The output floating point string is always one place less.
@bigdinotech Please review the code change.


bigdinotech:
Code looks good to me.
@calvinatintel please merge

PR #171, opened 2016-04-26
Jira-560, getInterruptStatus(...) should return type of bool

jgithubs:
Please review this change. For all features, all functions returns a 'bool', therefore, the main function should return a bool.

The team advise that I make the change and have your review/approve it.


jgithubs:
@sandeepmistry, please review/approve.

sandeepmistry:
@jgithubs looks good

PR #178, opened 2016-05-06
Bug fixed: CurieBle library, class BLECharacteristic allocates memory…

SidLeung:
… using an incorrect variable which may result in memory corruption if the memory size is bigger than 16 bytes.


SidLeung:
@bigdinotech Please review.

bigdinotech:
@calvinatintel looks good please merge

bbaltz505:
cherry-picked

PR #188, opened 2016-05-19
Jira-599.  Routine, dtostrf, did not work with the latest toolchain. …

SidLeung:
… The call to sprintf in the routine crashed the system.  Issue corrected by replacing the call.  The routine, printFloat, in Print.cpp, is now using dtostrf to print out double and float values.
Please note that dtostrf is almost a complete replacement.  From the AVR library reference description, the input parameter, width, can be negative - left spacing adjustment.

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -288,62 +288,8 @@ size_t Print::printLongLong(unsigned long long n, uint8_t base) {
 
 size_t Print::printFloat(double number, uint8_t digits)
 {
-  size_t n = 0;
-  int exponent = 0;
-  double tmp;
+  char str[50];
 
-  if (isnan(number)) return print("nan");
-  if (isinf(number)) return print("inf");
-
-  // Handle negative numbers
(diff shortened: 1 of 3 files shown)


SidLeung:
@bigdinotech @calvinatintel Please review the code.  It is desirable to have this change in the next release together with the new toolchain.

eriknyquist:
At first read this seems like we might be re-inventing the wheel. Instead of re-writing the sprintf functionality yourself, did you try replacing the sprintf calls with snprintf + explicit null-termination? i.e.
`snprintf(buf, size, ...);`
`buf[size - 1] = 0;`

From looking at how sprintf was being used in the old code (recklessly), it seems like the issue could have been cause by sprintf overflowing OR sprintf failing to null-terminate a string (I believe some implementations do not guarantee null-termination).

SidLeung:
@calvinatintel @eriknyquist Eirk said, "merge it".

calvinatintel:
Cherry-picked

PR #197, opened 2016-05-26
Jira-602: Convert float and double to strings.  Output string is padd…

SidLeung:
…ed with spaces.

The padded spaces in the output string is caused by the default value put as the (minimal) "width" parameter of the dtostrf call.  This parameter is set to 0 now.  In addition, the buffer allocated for constructing the output string is not fixed.  The buffer is allocated according to the size of the float/double when it is converted into a string (counting all digits, sign, and decimal).

### DIFF
--- a/cores/arduino/WString.cpp
+++ b/cores/arduino/WString.cpp
@@ -133,16 +133,24 @@ String::String(unsigned long long value, unsigned char base)
 
 String::String(float value, unsigned char decimalPlaces)
 {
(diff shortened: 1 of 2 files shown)


SidLeung:
@bigdinotech @eriknyquist @calvinatintel Please review the fix for this bug found in yesterday's BAT testing.

bigdinotech:
@calvinatintel looks good.

PR #202, opened 2016-05-31
ATLEDGE-586: String.compareTo() does not compute the same as Uno for POS values

calvinatintel:
@eriknyquist @bigdinotech @SidLeung Please review

### DIFF
--- a/cores/arduino/WString.cpp
+++ b/cores/arduino/WString.cpp
@@ -502,7 +502,13 @@ int String::compareTo(const String &s) const
 		if (buffer && len > 0) return *(unsigned char *)buffer;
 		return 0;
 	}
-	return strcmp(buffer, s.buffer);
+	const char *p1 = buffer;
+	const char *p2 = s.buffer;
+
+	while (*p1 == *p2++)
+		if ('\0' == *p1++)
+			return 0;
+	return (*(const char *)p1 - *(const char *)(p2 - 1));
 }
 
 unsigned char String::equals(const String &s2) const


eriknyquist:
@calvinatintel looks good. I reviewed and verified this fix last Friday (there is an example sketch attached to ATLEDGE-586 which can be used to verify, if anyone else needs to).

Summary: The libc++ that comes with our toolchain contains an implementation of strcmp which behaves differently from the behaviour described in the API reference (not to be confused with C strcmp from string.h, which has slightly different behaviour again). The String.compareTo function is using strcmp, so gives incorrect return values when using our toolchain.

But-- I must admit when I read the API ref for the compareTo function, I'm not too sure what it means.

reminder of C/C++ strcmp behaviour:

C++ strcmp-- positive return value means first non-matching char in str1 has greater value than same char in str2 (negative means lesser).
C strcmp -- positive return value means str1 is longer than str2 (negative means shorter)

And here's the reference for String.compareTo on the Arduino site:

a negative number: if string comes before string2
0: if string equals string2
a positive number: if string comes after string2

When i read this, it sounds like it could apply to the behaviour of both C and C++ strcmp.
It's a bit unclear. @bigdinotech @SidLeung what do you think? untimately we need to figure out what the public API reference actually means and follow that.

calvinatintel:
@bigdinotech @SidLeung Thumbs up? Down?

eriknyquist:
@SidLeung @bigdinotech what do you guys think? can you dechipher the meaning of the Arduino API ref. for String.compareTo?

eriknyquist:
I've asked for confirmation on basecamp
https://basecamp.com/2546761/projects/7617309/messages/58936653

eriknyquist:
OK, I think I'm happy now. The current behaviour aligns with the public Arduino API.
@calvinatintel please merge.

PR #207, opened 2016-06-03
SerialFlash: enlarge chip ID buffer

eriknyquist:
SerialFlashChip::readID() writes up to 5 bytes into
the buffer provided, but the caller only allocates
3 bytes. Increase buffer size to 5 bytes.

### DIFF
--- a/libraries/SerialFlash/SerialFlashDirectory.cpp
+++ b/libraries/SerialFlash/SerialFlashDirectory.cpp
@@ -310,7 +310,9 @@ bool SerialFlashChip::create(const char *filename, uint32_t length, uint32_t ali
 	// last check, if enough space exists...
 	len = strlen(filename);
 	// TODO: check for enough string space for filename
-	uint8_t id[3];
+
+    // 5 bytes, to allow for extra 2 bytes in Spansion device IDs
+	uint8_t id[5];
 	SerialFlash.readID(id);
 	if (address + length > SerialFlash.capacity(id)) return false;
 


eriknyquist:
@calvinatintel please don't merge yet-- I'd like to run a Klocwork scan first.

eriknyquist:
@SidLeung @bigdinotech ready for review

eriknyquist:
Some context; the offending code is only executed when a Spansion device is detected. The 101's onboard SPI Flash (the one which the SerialFlash library will use) is a Winbond, so the only way a user could see this bug while using a 101 is by modifying the SerialFlash library to use SPI (instead of SPI1) and connecting a Spansion device to it. In my opinion, this doesn't affect Bellatrix, what do you guys think?

bbaltz505:
Maybe just add a comment then.

SidLeung:
@eriknyquist It's not going to affect Bellatrix and the chances of having someone using SPI on a Spansion device is small.  Fix it, push it, and release it on the next round, Castor.  BTW, is there a reason between 3 and 5?  If so, would you please put it in as comment in the code too?  Thanks.

eriknyquist:
@bbaltz505 I was trying to suggest that this commit doesn't need to go into Bellatrix. Do you mean add a comment to this PR?

bbaltz505:
@eriknyquist , I meant if you were to touch the code, just add a comment. No need to reopen it for Bellatrix though - given your explanation.

eriknyquist:
I sent Paul a PR for the same.
https://github.com/PaulStoffregen/SerialFlash/pull/28

Since Paul's library works out of the box on a 101 now, I think we want to take it out of the next release and just direct users to download the latest version through the IDE library manager. So I'm not even sure if this PR should be merged?

eriknyquist:
@SidLeung mised your earlier comment, yes the reason for 5 is in the SerialFlashChip::readID function; if the device is a Spansion, then `buf[3]` and `buf[4]` are written, but in all other cases only the first 3 bytes are used.

eriknyquist:
No Klocwork issues in this PR, thanks for the help there @bbaltz505 :)

eriknyquist:
@calvinatintel ready to merge for Castor

calvinatintel:
cherry-picked

PaulStoffregen:
Also applied upstream :)  https://github.com/PaulStoffregen/SerialFlash/commit/05965b521802fdfbc6dccd15618ae564100dd0eb

eriknyquist:
Thanks @PaulStoffregen!

PR #209, opened 2016-06-08
Jira-617: String Object constructor hangs up when handling large floa…

SidLeung:
…ting point number

There is a bug in determining the number of digits before the decimal point of a floating point number.  The method employed used casting to rid of the decimal portion of the floating point value that resulted in an incorrect number of digits if the floating point value is large enough (overflow condition).

Casting removed.  Used another method to determine the number of digits in the integer portion of a floating point number (exponent count).

### DIFF
--- a/cores/arduino/WString.cpp
+++ b/cores/arduino/WString.cpp
@@ -133,23 +133,23 @@ String::String(unsigned long long value, unsigned char base)
 
 String::String(float value, unsigned char decimalPlaces)
 {
-	int len = digitsBe4Decimal(value);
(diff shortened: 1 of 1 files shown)


SidLeung:
@eriknyquist @bigdinotech Please review the changes.  @calvinatintel Please merge upon review approval.  @russmcinnis Please test the fix upon the generation of the new JSON.

bigdinotech:
@calvinatintel looks good to me. Please merge

eriknyquist:
@calvinatintel looks good

PR #210, opened 2016-06-10
Added setValue for strings

agdl:
This fix allows to use setValue() for const char\* variables

### DIFF
--- a/cores/arduino/Udp.h
+++ b/cores/arduino/Udp.h
@@ -41,7 +41,8 @@
 class UDP : public Stream {
 
 public:
-  virtual uint8_t begin(uint16_t) =0;	// initialize, start listening on specified port. Returns 1 if successful, 0 if there are no sockets available to use
+  virtual uint8_t begin(uint16_t) =0;  // initialize, start listening on specified port. Returns 1 if successful, 0 if there are no sockets available to use
+  virtual uint8_t beginMulticast(IPAddress, uint16_t) { return 0; }  // initialize, start listening on specified multicast IP address and port. Returns 1 if successful, 0 on failure
   virtual void stop() =0;  // Finish with the UDP socket
 
   // Sending UDP packets


kitsunami:
Can you provide the context for this?

agdl:
Yes of course:
```
BLECharacteristic stringCharacteristic("xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx", BLERead | BLEWrite | BLENotify, MAX_LENGTH);
...
...
...
void myTestFunction(){
  char testString[] = "This is a test";
  stringCharacteristic.setValue(testString);
}
```

or something like:

```
BLECharacteristic stringCharacteristic("xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx", BLERead | BLEWrite | BLENotify, MAX_LENGTH);
...
...
...
void myTestFunction(){
 String testString = "Value is: " + thisValue;
  stringCharacteristic.setValue(testString.c_str());
}
```

SidLeung:
The BLE library is undergo improvement.  Will add in this change as part of the modification.  Please note that "notification" is limited to 20 byte long.  If the string you are sending out is longer than 20 byte, you may have to send it out in 20 byte segments.  In that case, you can consider using an index pointer to walk down the string and specify the segment size for each setValue() call.  You definitely do not want to just call setValue( myLongString );

eriknyquist:
The change looks fine to me, but I'm not sure if it's a necessary one. @SidLeung the BLE master what do you think?

sandeepmistry:
This is one of the API's that I forgot to port from [BLEPeripheral](https://github.com/sandeepmistry/arduino-BLEPeripheral). It makes it much nicer for setting a string value for the characteristic vs `characteristic.setValue(str, strlen(str))`. @agdl noticed it and created this PR to port it over.

More info. at: https://github.com/sandeepmistry/arduino-BLEPeripheral/blob/master/API.md#set-value

SidLeung:
PR #461 addresses this issue.  Will be in the next release.  Please review and close.

PR #211, opened 2016-06-13
ATLEDGE-572: Fix getStepDetectionMode()

eriknyquist:
This function is reading all 8 bits of RA_STEP_CONF_1 (0x7A-0x7B)
to obtain a value for min_step_buf, though this value is only
represented by the lower 3 bits; bit 4 is used as the 'enable' bit.
As a consequence, this function will read the step detection mode
incorrectly when step counting is enabled. Fix this by only reading
the lower 3 bits of RA_STEP_CONF_1.


eriknyquist:
@SidLeung @bigdinotech @calvinatintel ready for review

bigdinotech:
@calvinatintel looks good please merge

PR #217, opened 2016-06-17
ATLEDGE-619: fix dtostrf

eriknyquist:
Use a different method for float->string conversion, and
fix the width calculation so that extra spaces are only
printed if the width parameter exceeds the output string
length (ie. a mininum width, as specified by dtostrf
reference from avr-libc).

### DIFF
--- a/cores/arduino/WString.cpp
+++ b/cores/arduino/WString.cpp
@@ -813,26 +813,3 @@ float String::toFloat(void) const
         if (buffer) return float(atof(buffer));
         return 0;
 }
-
-/*********************************************/
-/*  utilities functions                      */
-/*********************************************/
-
-int String::digitsBe4Decimal(double number)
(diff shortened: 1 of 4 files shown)


eriknyquist:
Commit message is self explanatory, except for the fact that this change also makes the test values from ATLEDGE-619 print correctly. @calvinatintel could we get a JSON for this before merging so Janet can do some further testing?

bbaltz505:
@eriknyquist , try http://mkfs.ndg.intel.com/precommit/json/package_0.0.26_index.json

SidLeung:
@eriknyquist  Looks good.  @bigdinotech Please check it out.

eriknyquist:
@calvinatintel ready to merge.

bbaltz505:
cherry-picked

PR #218, opened 2016-06-20
JIRA-624: I2C slave mode

xieqi:
added I2C controller slave mode support for Lakemont.

### DIFF
--- a/cores/arduino/soc_i2c.c
+++ b/cores/arduino/soc_i2c.c
@@ -135,7 +135,7 @@ static int soc_i2c_wait_dev_ready(SOC_I2C_CONTROLLER controller_id,
     return I2C_TIMEOUT - ret;
 }
 
-int soc_i2c_openadapter(uint32_t address, int i2c_speed, int i2c_addr_mode)
+int soc_i2c_open_adapter(uint32_t address, int i2c_speed, int i2c_addr_mode)
 {
     int ret = 0;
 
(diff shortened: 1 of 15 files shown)


xieqi:
@SidLeung @linrjing @sgbhu
lakemont I2C controller driver, now it only finished master mode, because no I2C pin export, so just use bus scan sketch to verify the driver framework, will add slave mode support after we got Radon board.

cores/arduino/soc_i2c.c
```diff
@@ -0,0 +1,246 @@
...
+			{
+				return I2C_ERROR_OTHER; // other error
+			}
+		}
+		if (!no_stop) {
+			if (soc_i2c_tx_complete) {
+			return I2C_OK;
+			}
+		}
+		delay(1);
```
eriknyquist: I see a lot of loops that have been "spaced out" using delay(1). That's fine, it shouldn't have any negative effects since I2C is quite slow anyway, but I don't think it will gain anything either. Remember, this is for a single-threaded MCU device.

gpkc:
How is this going to work without a onReceive implementation?

SidLeung:
Please note that the physical I2C slave mode interface is **not** available on the Arduino 101 hardware platform and, thus, it is not going to be part of Corelibs.  This is done as a preparation for future products.

gpkc:
Yes, I'm aware of that. I thought this was a software implementation or something like this. I'm not sure if this would be possible, though.

xieqi:
@SidLeung @eriknyquist @linrjing 
update the code to full support I2C slave mode, please review the code, thanks.

SidLeung:
Please do **not** make any additional changes to this branch until further notice.

SidLeung:
@bbaltz505 Please close this branch without merging it to the main trunk.

PR #237, opened 2016-07-12
define digitalPinToInterrupt for backwards compatibility

facchinm:
see http://forum.arduino.cc/index.php?topic=370167.0


calvinatintel:
cherry-picked

PR #240, opened 2016-07-15
Jira566: I2S DMA implement: supplement library and example

linrjing:
@SidLeung @calvinatintel@xieqi @sgbhu update the I2S DMA implement codes:(1) work well on different sample resolution(12,16,24,32 bits);(2)  Put a time delay in the interrupt to make clock stay longer to make sure all data have been arrived RX side. And don't need to push extra zero to register at TX side any more.


bigdinotech:
@calvinatintel @linrjing looks good +1

calvinatintel:
cherry-picked

Issue #149, opened 2016-03-15
Please support calibrated accel & gyro in CurieIMU's API


PaulStoffregen:
Please consider adding APIs for Arduino users to get properly scaled data, with the conversion factor done inside the CurieIMU library.  There are at least two reasons why I hope you'll do this:

1) The library can "know" which range is in use, to automatically perform the scaling correctly.  In the examples, like **File > Examples > CurieIMU > Accelerometer**, if a novice Arduino user edits [line 36](https://github.com/01org/corelibs-arduino101/blob/master/libraries/CurieIMU/examples/Accelerometer/Accelerometer.ino#L36) to configure a wider range, they must also know/remember to edit [line 69](https://github.com/01org/corelibs-arduino101/blob/master/libraries/CurieIMU/examples/Accelerometer/Accelerometer.ino#L69) to agree with their range setting.  While this isn't very difficult for experienced programmers, reducing the possibility for things to go wrong, to make everything "just work" as simply as possible is the design style that makes Arduino so successful.

2) Over the long-term, this API is likely to be used with other sensors which are likely to have different scaling.  By abstracting these hardware specific details and avoiding them within the user's code, the API can be much more useful over the very long term.

Probably the simplest approach would be functions with the same names which take references to float, and populate the floats with data properly scaled.

eriknyquist:
Hi @PaulStoffregen, so currently raw accelerometer data (for example) is grabbed with
`void CurieIMUClass::readAccelerometer(int& x, int& y, int& z)`. If we added another function like this
`void CurieIMUClass::readAccelerometer(int& x, int& y, int& z, int range)`, where 'range' is the same value passed to `setAccelerometerRange` and the resulting values are scaled accordingly.
Is this what you mean?

PR #243, opened 2016-07-18
Issue #149: Provide scaled IMU data

eriknyquist:
Add implementations of readAccelerometer and readGyro which
return values that are scaled according to the range set with
setAccelerometerRange and setGyroRange


eriknyquist:
@bigdinotech @SidLeung @calvinatintel ready for review.

bigdinotech:
@calvinatintel @eriknyquist +1

calvinatintel:
cherry-picked

PR #245, opened 2016-07-20
ATLEDGE-643: Add chip ID to SerialFlash example sketch

eriknyquist:
RawHardwareTest reports the 101's on-board SPI flash chip as an unknown
device. Fix this by adding the device's JEDEC ID to the example sketch.


eriknyquist:
@bigdinotech @SidLeung @calvinatintel ready for review

bigdinotech:
@eriknyquist @calvinatintel looks good +1

PR #250, opened 2016-07-27
Add delay to Serial1.end() allow GPIO lines to settle

eriknyquist:
Through testing, it was discovered that if Serial1.begin() is called again
immediately after Serial1.end(), then any attempt to write data immediately
following the begin() call could result in corrupted frames.

After calling Serial1.end(), it will be at least 300us before those pins can
be used as UART again.

### DIFF
--- a/cores/arduino/UARTClass.cpp
+++ b/cores/arduino/UARTClass.cpp
@@ -25,6 +25,8 @@
 #include "wiring_constants.h"
 #include "wiring_digital.h"
 
+#define SETTLING_TIME  400
+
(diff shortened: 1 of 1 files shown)


eriknyquist:
@bigdinotech @SidLeung @calvinatintel please review, but do not merge until I confirm test results from Russ

bigdinotech:
@eriknyquist @calvinatintel looks good to me

calvinatintel:
cherry-picked

eriknyquist:
@calvinatintel please waiting for testing confirmation next time

PR #259, opened 2016-08-03
Fix serial flush

descampsa:
Currently, flush wait for the transmission to be complete by checking if the transmission holding register (or fifo) is empty.
When this happen, the last byte is still in the shift register, and has still to be transmitted. This mean we can't know when the transmission is actually complete, which is crucial to implement RS485 communication, for example.
This pull request change this behaviour and wait for the shift register to be empty, which means the transmission is really complete. 
This new behaviour is  also more consistent with the behaviour of flus on AVR boards.

Note that the change have not be extensively tested, and i am not sure this is entirely correct, so it would be good if someone with a better knowledge of the uart device could have a look.

### DIFF
--- a/cores/arduino/UARTClass.cpp
+++ b/cores/arduino/UARTClass.cpp
@@ -169,7 +169,7 @@ void UARTClass::flush( void )
 {
   while (_tx_buffer->_iHead != _tx_buffer->_iTail); //wait for transmit data to be sent
   // Wait for transmission to complete
-  while (uart_irq_tx_ready(CONFIG_UART_CONSOLE_INDEX));
+  while(!uart_tx_complete(CONFIG_UART_CONSOLE_INDEX));
 }
 
 size_t UARTClass::write( const uint8_t uc_data )
--- a/system/libarc32_arduino101/drivers/ns16550.c
+++ b/system/libarc32_arduino101/drivers/ns16550.c
@@ -626,3 +626,16 @@ void uart_int_connect(int which,	   /* UART to which to connect */
 	/* set the Host Processor Interrupt Routing Mask */
 	SOC_UNMASK_INTERRUPTS(INT_UART_0_MASK + (which * UART_REG_ADDR_INTERVAL));
 }
+
+/*******************************************************************************
+*
+* uart_tx_complete - check if tx holding and shift register are empty
+*
+* RETURNS: zero if registers are non-empty (transmission not complete), 
+*          non-zero if registers are empty (transmission complete)
+*/
+
+uint8_t uart_tx_complete(int which)
+{
+	return INBYTE(LSR(which)) & LSR_TEMT;
+}
--- a/system/libarc32_arduino101/drivers/uart.h
+++ b/system/libarc32_arduino101/drivers/uart.h
@@ -91,6 +91,7 @@ int uart_line_status(int port);
 int uart_break_check(int port);
 void uart_break_send(int port, int delay);
 void uart_disable(int port);
+uint8_t uart_tx_complete(int which);
 
 #ifdef __cplusplus
 }


ndgbuilder:
Hello, and thanks for your contribution! Unfortunately, your commit message does not follow our [guidelines for considerate commit messages](https://github.com/01org/corelibs-arduino101/wiki/Writing-a-commit-message).
Specifically:

* Lines in description exceed 80 characters

Please fix these issues by amending the commit (`git commit --amend`).

eriknyquist:
Hi @descampsa, can you squash these three commits into one please? currently each individual commit would not compile on its own.

descampsa:
I have merged the three commits, and modified the commit message to respect the 80 characters limit.

eriknyquist:
Thanks @descampsa. The change looks OK, I just have two suggestions/questions;

1. Do you have some code  that shows a problem with Serial.flush(), and which no longer has that problem after applying your fix?

2. I think it might make more sense to put your uart_tx_complete function into the driver with the rest of the ns16550 routines, in system/libarc32_arduino101/drivers/ns16550.c, and to rename it so it matches the rest of the routines, e.g. `uart_irq_tx_complete`. What do you think?

descampsa:
1. I noticed this problem when trying to implement a half duplex communication with a dynamixel motor, but i don't have anything fully functionnal yet, and anyway it is unlikely that you have such a device to test.
However, you can test it easily by reversing the commit https://github.com/01org/corelibs-arduino101/commit/9a871b671ceec5f625c63c9a9e58f044cef2e70c. Just send some data, close the connection, reopen it and send some data again. The second data should be corrupted without my fix, and correct with it (i tested with a baud rate of 9600).
Actually, i think that this commit was just a workaround for the problem described here, and it won't be necessary anymore. This requires more testing to be sure, however, especially to check that the behaviour is consistent at different baudrates.

2. I don't understand, the function is already in system/libarc32_arduino101/drivers/ns16550.c, right?
As for renaming it, i don't think `uart_irq_tx_complete` is appropriate, because all `*irq*` function are based on the interrupt state (Interrupt Ident Register), while this function use the Line Status Register.

eriknyquist:
1. Can you please provide a sketch which demonstrates the problem
2. Yeah you're right, it's already in there, my mistake. And your naming makes sense after you explained it.

eriknyquist:
@descampsa I tried out some tests with your change and the I see no change in behaviour.
I still get corrupted data. This is why you need to provide sample code; if we're not testing the same thing then things get difficult. Can you share some please?

descampsa:
Here is a first sample code : 
[test_dyn_led.ino.zip](https://github.com/01org/corelibs-arduino101/files/400369/test_dyn_led.ino.zip)

This sample case is similar to https://github.com/01org/corelibs-arduino101/commit/9a871b671ceec5f625c63c9a9e58f044cef2e70c, and  in my test, my change fixes the issue.

descampsa:
Ok, here is a better example: 
[test_dyn_led.ino.txt](https://github.com/01org/corelibs-arduino101/files/400534/test_dyn_led.ino.txt)

If you monitor the pin 1 with a logic analyzer, you will see that, without my change, the last byte is not sent, and that my change fix that.
nonfixed
fixed


Again, this is tested without https://github.com/01org/corelibs-arduino101/commit/9a871b671ceec5f625c63c9a9e58f044cef2e70c, because this commit is a (ugly) workarround of this problem.

descampsa:
Here is another example, that appears even if https://github.com/01org/corelibs-arduino101/commit/9a871b671ceec5f625c63c9a9e58f044cef2e70c is applied:
[test_dyn_led.ino.txt](https://github.com/01org/corelibs-arduino101/files/400604/test_dyn_led.ino.txt)

This is an example of how a half duplex communication would be implemented : a direction pin switch between transmit and receive state (it would be connected to the DE and RE* ports of a ds485 for example). In this case, the arduino send a command and then wait for a response from a slave device, so he direction pin must switch from transmit to receive immediatly after the last byte has been transmitted, and not before. This is currently not possible to implement without my change.

Before (channel 1 is direction pin):
nonfixed
After:
fixed

eriknyquist:
Thanks for all the info @descampsa. You are right, that workaround is an ugly one, but it's actually a workaround for slightly different problem that we encountered with some of our Serial tests, and your fix doesn't seem to change the behaviour in that particular case (this is the "corrupted data" I spoke of before I had your sample code).

I think your change makes sense though, based on what the HW specs say... I'll have some time to test these things out tomorrow and I'll let you know then.

descampsa:
Ok, no problem, from my tests i actually thought that those problems were the same, and that my change would fix both issue, but i was wrong apparently. That was hard to be sure without a complete datasheet of the microcontroler ;-)
Out of curiosity, would you mind to share your test code for this issue?

russmcinnis:
Hey descampsa. Here is a sketch that reproduces the issue Erik and I are seeing.
[Serial1_error.zip](https://github.com/01org/corelibs-arduino101/files/401925/Serial1_error.zip)

I have tried with and without the Serial1.flush() and get the same results.
[Serial1_error.zip](https://github.com/01org/corelibs-arduino101/files/401938/Serial1_error.zip)



Attached serial1_error.ino which reproduces the issue without any changes. If I uncomment the lines where the baud rate is printed there is no failure. 
// Serial.print("Baud:");
// Serial.println(baudRateTestList[i]);
Or is you comment out the Serial1.end() there is not failure. The sketch requires pins D0-D1 to be jumpered. Open serial terminal at 115200. The test cycles through the baud rates using Serial1.print() and Serial1.read().
Put a delay of 600 microseconds after Serial1.end() and the test passes. Change the delay to 500 and it fails.
// Reset for next loop
Serial1.end();
delayMicroseconds(500);

descampsa:
Hey russmcinnis, thanks for the test code. 
I have tested your code, and indeed it seems that it is a completely different, unrelated, problem, but it appears in a similar context, hence the confusion.
I am still testing it, but i think i have found the root cause of that problem, and maybe a possible fix.
I am going to open a new issue for that problem with a detailled description when i have some time, probably this evening or tomorrow.

eriknyquist:
Hi @descampsa. I'm looking at this problem today as well, and I must confess I'm a bit stumped. Your fix only checks 1) for the SRAM buffer to be empty, and 2) for the shift register to be empty.

So I was quite sure that the right thing to do is 1) wait for SRAM buffer to be empty, 2) wait for THR to be empty, and finally 3) wait for shift register to be empty. 

This didn't fix it, though. Still, the only way that I can get around this is by putting that ugly delay in Serial.end():

    SET_PIN_MODE(17, GPIO_MUX_MODE); // Rdx SOC PIN (Arduino header pin 0)
    SET_PIN_MODE(16, GPIO_MUX_MODE); // Txd SOC PIN (Arduino header pin 1)
    delayMicroseconds(400);

I thought maybe this was just the lines settling after the signals change? but that can't be it, 400us is far too long for that. Pointers appreciated!

descampsa:
Hey,
To give you a pointer, i think that the problem is that when the RX pin is
switched to gpio, the input of the uart module goes to low level. If the
duration between end and begin is long enough, it is considered as a line
break, and if it is short enough, no packet start is detected. However, if
it is between the two, it is considered as a packet, and, worst, if you
send a real packet fast enough, it can be seen as part of the false packet,
and the state of the receiver is then completely wrong, hence the data
corruption. I have a test case to prove that, with the right timing, you
may receive that false packet without ever sending anything, i will post it
tomorow but i dont have it at hand right now.

One simple solution would be to stop the uart completely, but i don t know
if it is possible. Another option could be to enable the loopback, to keep
the input level at a high level, but i have not tested it.

descampsa:
@eriknyquist Yeah, i was also quite puzzled when i saw the result of my tests, and that got me thinking for a while before i understood....
Anyway, i have opened a separate pull request, since this is a completely independent problem from this one.

descampsa:
About the THR, i don't think it is necessary to wait for it explicitely, because TI [datasheet,](http://www.ti.com/lit/ds/symlink/pc16550d.pdf) for example, says that 
> Bit 6 is set to a logic 1 whenever the Transmitter Holding Regis- ter (THR) and the Transmitter Shift Register (TSR) are both empty. It is reset to a logic 0 whenever either the THR or TSR contains a data character. In the FIFO mode this bit is set to one whenever the transmitter FIFO and shift register are both empty.

If this quote is valid in this context, waiting for TEMT is enough to ensure that both THR and the shift register are empty.

eriknyquist:
@descampsa yes, I agree. Now that you have also fixed PR #260 ( and we know that they are separate issues), I think this can be merged now @calvinatintel

calvinatintel:
cherry-picked

PR #264, opened 2016-08-10
JIRA-668: Removing Intel version of Adafruit_NeoPixel

bbaltz505:


eriknyquist:
@bbaltz has anyone verified that the standard Neopixel library works on a 101?

russmcinnis:
I verified that NeoPixel is working with released FW/corelibs. I deleted the adafruit neopixel library from arduino15 and then used the library manager to get the 1.0.6 version from adafruit. I let the strand test run for over an hour with a 60 light strand.

calvinatintel:
Yeah, we test our NeoPixel libraries as a part of BAT. I'm surprised that there's a problem. I'll check too and report back

eriknyquist:
@calvinatintel I mean the Adafruit Neopixel library, not our one. Apparently the Adafruit one supports the 101 now.

PR #269, opened 2016-08-17
Add flashpack to IDE

calvinatintel:
Needs to be paired with a json which installs flashpack.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -28,6 +28,8 @@ arduino_101.build.variant_system_lib=arc32drv_arduino101
 arduino_101.build.vid=0x8087
 arduino_101.build.pid=0x0AB6
 
+arduino_101.bootloader.tool=arduino101load
+
 ##############################################################
 
 
--- a/platform.txt
+++ b/platform.txt
@@ -5,7 +5,7 @@
 # https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5---3rd-party-Hardware-specification
 
 name=Intel Curie (32-bit) Boards
-version=1.0.6
+version=1.0.7
 
 # Arduino 101 compile variables
 # ----------------------
@@ -99,3 +99,12 @@ tools.arduino101load.upload.params.verbose=verbose
 tools.arduino101load.upload.params.quiet=quiet
 
 tools.arduino101load.upload.pattern="{cmd.path}" "{runtime.tools.arduino101load.path}/x86/bin" {build.path}/{build.project_name}.bin {serial.port} "{upload.verbose}"
+
+# This is needed to avoid an error on unexistent fields
+tools.arduino101load.erase.params.verbose=
+tools.arduino101load.erase.params.quiet=
+tools.arduino101load.erase.pattern=
+tools.arduino101load.bootloader.params.verbose=
+tools.arduino101load.bootloader.params.quiet=
+tools.arduino101load.bootloader.pattern={runtime.tools.flashpack.path}/flash_dfu.sh
+tools.arduino101load.bootloader.pattern.windows={runtime.tools.flashpack.path}/flash_dfu.bat
--- a/programmers.txt
+++ b/programmers.txt
@@ -1,0 +1,6 @@
+arduino101fw.name=Arduino/Genuino 101 Firmware Updater
+arduino101fw.communication=USB
+arduino101fw.protocol=
+arduino101fw.program.protocol=
+arduino101fw.program.tool=arduino101load
+arduino101fw.program.extra_params=


calvinatintel:
My [first attempt at the json](http://mkfs.ndg.intel.com/dev/json/package_0.1.53_index.json)

1. Installs the factory recovery flashpack
2. Populates **Programmer** entry
3. Triggers `flash_dfu.sh` upon clicking **Burn Bootloader** button

screen shot 2016-08-17 at 11 36 27 am

However, the flashpack hangs after the board is reset, and does not initiate flashing.
screen shot 2016-08-17 at 11 39 40 am

Considering that the board reappears on **Tools > Port** menu within 5 seconds, it seems that `flash_dfu.sh` script didn't catch the board in DFU mode. Running the script manually from within `.arduino15` directory correctly executed the script. Perhaps we need the baudrate reset mechanism built into the flashpack.

calvinatintel:
A couple tasks left:

1. Reduce flashpack to BLE only. Maybe we can attach the `factory-ble-flashpack.tar.bz2` to Castor release and link it back to json? So meta.
2. Try to fix the hanging issue
    - Test on other OSes to see if this is an OSX quirk wrt quanrantine bit
    - Add the baudrate reset script into the flashpack
    - Make a wrapper script which calls the baudrate reset script, followed by `flash_dfu.sh`

calvinatintel:
First attempt at [arduino101-factory_ble-flashpack.tar.bz2](http://mkfs.ndg.intel.com/jenkins_artifacts/pvt/arduino101-factory_ble-flashpack.tar.bz2).
Works on OSX. Need other OSes tested.

eriknyquist:
Nice. will verify after kanban.

bbaltz505:
Tested programmer on Windows. I had to add a Windows-specific instruction in platform.txt:
`tools.arduino101load.bootloader.pattern.windows=cmd /C {runtime.tools.flashpack.path}/flash_dfu.bat`
but it also seemed to hang

noelpaz:
You verify this by using a BLE Central corelibs . we need to be testing the Central Examples. Scanning should scan for available BLE devices

noelpaz:
Will test with Brian's fixes on windows and Ubuntu

platform.txt
```diff
@@ -99,3 +99,11 @@ tools.arduino101load.upload.params.verbose=verbose
...
 
 tools.arduino101load.upload.pattern="{cmd.path}" "{runtime.tools.arduino101load.path}/x86/bin" {build.path}/{build.project_name}.bin {serial.port} "{upload.verbose}"
+
+# This is needed to avoid an error on unexistent fields
+tools.arduino101load.erase.params.verbose=
+tools.arduino101load.erase.params.quiet=
+tools.arduino101load.erase.pattern=
+tools.arduino101load.bootloader.params.verbose=
+tools.arduino101load.bootloader.params.quiet=
+tools.arduino101load.bootloader.pattern=/bin/bash {runtime.tools.flashpack.path}/flash_dfu.sh
```
calvinatintel: We need to make a version of this which works for both Linux, OSX, and Windows. Currently only works in Linux and OSX.

bigdinotech: If you look at how we did it in Edison you should see how we used a different method for each OS

bbaltz505: I pushed a fix for Windows. The other OSes should support bash.

calvinatintel: Awesome. Can you guys please test it?

boards.txt
```diff
@@ -28,6 +28,9 @@ arduino_101.build.variant_system_lib=arc32drv_arduino101
 arduino_101.build.vid=0x8087
 arduino_101.build.pid=0x0AB6
 
+arduino_101.bootloader.tool=arduino101load
+arduino_101.bootloader.file=""
```
bbaltz505: This causes the IDE to give a warning:
`Bootloader file specified but missing: /home/babaltz/.arduino15/packages/Intel-Test/hardware/arc32/0.0.49/bootloaders/""`

calvinatintel: I saw that too. Since that file isn't used, perhaps we can point that to the actual `image.bin` to suppress the error?

bbaltz505: Removing that line seems to resolve the warning

calvinatintel: Alright, let's go with that

calvinatintel:
I think we have a working solution. I'll publish a new json in 10 min.

calvinatintel:
http://mkfs.ndg.intel.com/precommit/json/package_0.0.50_index.json

Works on OSX. Please verify on Linux and Windows.
If this works on all OSes, we're ready for 1.0.7.

calvinatintel:
Verified on Linux 64 and Linux 32.

PR #274, opened 2016-08-27
Serial discard line break

descampsa:
Enable the Receiver Line Status interrupt source in order to discard the 0x00 byte that is generated in case of line break.

I need this because i use the uart to implement a half duplex communication by connecting rx and tx together and enabling/disabling rx/tx. When i disable rx, a line break is generated, and a false data byte of 0x00 is read.

Maybe data shoud also be discarded in case of parity error and framing error? 

This code  is based on https://github.com/01org/corelibs-arduino101/pull/266

Test case : 
[serial3_error.ino.txt](https://github.com/01org/corelibs-arduino101/files/440838/serial3_error.ino.txt)

### DIFF
--- a/cores/arduino/UARTClass.cpp
+++ b/cores/arduino/UARTClass.cpp
@@ -218,9 +218,12 @@ void UARTClass::IrqHandler( void )
   // if irq is Transmitter Holding Register
   else if(uart_irq_tx_ready(CONFIG_UART_CONSOLE_INDEX))
   {
-    if (_tx_buffer->_iTail != _tx_buffer->_iHead) {
-      uart_poll_out(CONFIG_UART_CONSOLE_INDEX, _tx_buffer->_aucBuffer[_tx_buffer->_iTail]);
-      _tx_buffer->_iTail = (unsigned int)(_tx_buffer->_iTail + 1) % SERIAL_BUFFER_SIZE;
+    if(_tx_buffer->_iTail != _tx_buffer->_iHead)
+    {
+      int end = (_tx_buffer->_iTail < _tx_buffer->_iHead) ? _tx_buffer->_iHead:SERIAL_BUFFER_SIZE;
+      int l = min(end - _tx_buffer->_iTail, UART_FIFO_SIZE);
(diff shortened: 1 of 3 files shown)


descampsa:
I also have a question about the hardware implementation of the uart. 

From my experiments, it seems that when i disconnect the uart rx pin (using `SET_PIN_MODE(17, GPIO_MUX_MODE)`), it is consistently pulled down to gnd level (hence the line break). I assume that a pull down resistor, or a direct connection to gnd is connected to the rx pin?
Can you confirm that it is the case, and that there is no chance to receive false data in this case, due to em noise or whatever?

cores/arduino/UARTClass.cpp
```diff
@@ -168,7 +168,7 @@ int UARTClass::read( void )
 
 void UARTClass::flush( void )
 {
-  while (_tx_buffer->_iHead != _tx_buffer->_iTail); //wait for transmit data to be sent
+  while (_tx_buffer->_iHead != *(volatile int*)&(_tx_buffer->_iTail)); //wait for transmit data to be sent
```
SidLeung: Is there a particular reason for casting the pointer to a volatile pointer first prior to using it?  Is the code being optimized out by the compiler if the volatile casting is omitted?

descampsa: The reason is that iTail can be updated in the irq handler. 
Without the volatile, the compiler may optimize the code by reading it only once from memory, and keeping its value in register. That would lead to an infinite loop.

I don't remember if i actually ran into this problem or not, and i did not check the resulting assembly, but that look like a typical case where volatile is needed.

Note that this more related to pull request https://github.com/01org/corelibs-arduino101/pull/266 than to https://github.com/01org/corelibs-arduino101/pull/274

descampsa:
I will rebase this against https://github.com/01org/corelibs-arduino101/pull/266 (when merged), and complete it with the parity and framing error cases, so put it on hold for now.

novashah:
Released and merged with 266.

PR #275, opened 2016-08-29
Update expected BLE version for V3

bbaltz505:
Update expected BLE version with one found in V3 image.
Update the corelibs version to 2.0.0


eriknyquist:
+1

PR #277, opened 2016-09-01
Add CurieMailbox library

eriknyquist:
Zephyr sample applications to go with the example sketches
can be found in CODK-M-X86-Samples

### DIFF
--- a/libraries/CurieMailbox/README.txt
+++ b/libraries/CurieMailbox/README.txt
@@ -1,0 +1,111 @@
+CurieMailbox library for Arduino/Genuino 101
+
+Overview:
+
+This library provides access to the inter-processor mailbox that allows
+interrupt-based communication between the x86 and ARC cores contained in the
+Curie module (in such a transaction, the message sender is referred to as the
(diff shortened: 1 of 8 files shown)


eriknyquist:
@bigdinotech @SidLeung @calvinatintel please review

bigdinotech:
+1 looks good to me but I think this should go into the codk-m branch

eriknyquist:
That's what I thought at first too, but I also think it makes sense to avoid having "CODK-M only" libraries wherever possible. @calvinatintel @SidLeung  what do you think?

calvinatintel:
Well, if it only works on M, I guess it would make sense to put it in M branch https://github.com/01org/corelibs-arduino101/tree/codk-m

eriknyquist:
well you can't say "it only works on M". Right now, CODK-M is the only thing that we provide which makes sense to use with this library. It works just fine in the IDE, and although it would be an odd choice right now, I see no reason to restrict that.

eriknyquist:
As per our discussion last week, I will close this PR and make a new one for the M branch

PR #281, opened 2016-09-05
CurieEEPROM refactor

facchinm:
This PR refactors CurieEEPROM library, syncing it with AVR version.
Since EEPROM on Curie chip is based on OTP flash storage, the minimum write size is 4 bytes. The current implementation brings some problems like http://forum.arduino.cc/index.php?topic=421348.0 , and the examples are quite broken.

The refactor also changes the name, from CurieEEPROM to EEPROM, so code for AVR sketches compiles out of the box. The new library resolution engine (from IDE 1.6.8 onward) handles this perfectly but we can stick to older name if we need compatibility with older IDEs.


zeptochain:
This fix passes my local tests :) I had been struggling for many hours with CurieEEPROM 1.0.6

libraries/EEPROM/src/EEPROM.h
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  uint32_t currentDword = CurieRead32(address);
+
+  int offset = address%4;
+
+  uint32_t data32 = (currentDword & ~(uint32_t)(0xFF << ((3-offset)*8)));
+  data32 = data32 | (data << ((3-offset)*8));
+
+  if (currentDword != 0xFFFFFFFF) {
```
bigdinotech: maybe instead of checking the value of CurieRead32 to be 0xFFFFFFFF
check the value of CurieRead8 to be 0xFF instead?
That way we are not clearing and restoring the entire block 4x as much as before.

You can take a look at the implementation of CurieEEPROM::write8
I basically read the dword and insert the byte inside the dword depending on where it needs to be.

facchinm: Great! I believe I misinterpreted the datasheet, so I thought that the whole dword needed to be erased before rewriting. Testing with the "reduced" version works too in my test sketches. I'm pushing the modification right now.
@zeptochain , could you give a try to the updated patch with your test sketches? Thanks

zeptochain: @facchinm will try this over the weekend and report back

zeptochain: just looked at the commit - not really confident, but will try it - i have a feeling that Martino got it right first time - i'll first write further test sketches to prove out 281 before trying 291 for comparison...

zeptochain: OK take it all back -- this update passes my test sketches including a complex config job I needed this library to accomplish. Has my thumbs up @facchinm

calvinatintel:
Cherry-picked into the 1.0.7 branch. Will be closed once 1.0.7 is final.

calvinatintel:
cherry-picked

PR #284, opened 2016-09-07
Improve CurieBLE library

biagiom:
- Improve sketches that use Serial communication
- Update keyword.txt file of CurieBLE library
- Improve the BLEAttribute Class

### DIFF
--- a/libraries/CurieBLE/examples/BatteryMonitor/BatteryMonitor.ino
+++ b/libraries/CurieBLE/examples/BatteryMonitor/BatteryMonitor.ino
@@ -23,7 +23,8 @@ int oldBatteryLevel = 0;  // last battery level reading from analog input
 long previousMillis = 0;  // last time the battery level was checked, in ms
 
 void setup() {
-  Serial.begin(9600);    // initialize serial communication
+  Serial.begin(9600);    // initialize Serial communication
+  while(!Serial) ;       // wait for serial port to connect
   pinMode(13, OUTPUT);   // initialize the LED on pin 13 to indicate when a central is connected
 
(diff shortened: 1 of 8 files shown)


sandeepmistry:
We should be careful about adding `while(!Serial) ;` to the example sketches. As this forces people to open the serial monitor when using them. This also means the examples won't run when powering the 101 via a battery or USB wall charger.

The other changes look good to me though.

biagiom:
Hi @sandeepmistry,
thanks for your comments and I think you are right about to be careful when adding `while(!Serial) ;` since people have to open the Serial monitor to execute the next lines of code but I have added it since the sketches use the Serial communication and print some text in `void setup()`, but text could be lost if you don't add the line `while(!Serial) ;`.
I have also noticed that if you leave the Serial monitor open when uploading a new sketch to Arduino/Genuino 101 you can't use the Serial monitor during the upload procedure as well as some seconds after it, but in this way, you avoid the use of `while(!Serial) ;` since if you want to print some text with `Serial.print()` or `Serial.println()`in `void setup()` the text is not lost because when the Serial port is open the Serial monitor is already open. This prevent the loss of data send to the Serial monitor.
If you think that `while(!Serial) ;` is not a really thing to add, so I agree with you and the commits that add `while(!Serial) ;` could be skipped.
I hope to be clear.
Best regards and thank you very much for your contribution about the development of the great CurieBLE library (as well as Node and Bleno).

eriknyquist:
@biagiom please remove the unrelated commit 	b329d08

eriknyquist:
@biagiom please resolve merge conflict. @SidLeung please review.

PR #287, opened 2016-09-07
Improve sketches of Wire library for Arduino/Genuino 101

biagiom:
Pull request to improve sketches of Wire library for Arduino/Genuino 101 that use Serial communication


eriknyquist:
I don't think these changes are necessary. All the Wire examples run in an infinite loop, so you will always see output no matter when you open the Serial monitor.

biagiom:
Hi @eriknyquist,
I agree with you since in this case wesend data to the Serial Monitor only in `void loop()`.
I have updated the bus_scan.ino sketch of Wire library and I sent my changes with #296 Pull request

PR #290, opened 2016-09-09
reduce sahred memory buffer size

bigdinotech:
-reduce the buffer size used for the Shared Memory Communication library


bigdinotech:
@calvinatintel please merge
this is the missing file from the SMC fixes commit

PR #303, opened 2016-09-28
Rethink SerialFlash class

facchinm:
Removing the fork of SerialFlash library allows the freedom to track mainline without any maintenance burden. 
Since the library in no longer included, all examples have been patched to add a direct link to library manager.

_Do not merge yet_ : https://github.com/PaulStoffregen/SerialFlash/pull/35 needs to be merged and published before this patch applies correctly


eriknyquist:
Looks good @facchinm, I'll watch the PR on Paul's library and merge this one after it.
Just one suggestion, if there are only 2 differences (the overloaded `begin` and the CS pin) between these examples and Paul's examples, then I think it's better just to add those lines as comments in Paul's examples with a "use this line for Arduino/Genuino 101" message. Might as well avoid duplicating the example code too if we can!

facchinm:
I don't think it is a good idea to remove completely the examples, the OOTB experience could suffer (I agree that they are unnecessary since avoiding the overloaded constructor approach makes this library a pure wrapper). I commented on Paul's PR about tested compatibility so I'm confident it will be merged soon

sandeepmistry:
It would be awesome if we could change `SerialFlash.begin(SPI1, FlashChipSelect);` in the examples to something like `CurieSerialFlash.begin()` later on.

eriknyquist:
@sandeepmistry then, in order to convert a sketch from using default SPI to using 101 built in SPI-flash, you have to search/replace all occurrences of 'SerialFlash'. This way, you only have to change the `begin()` call.

sandeepmistry:
@eriknyquist good point regarding re-using existing sketches.

I'm still not too keen on making the user's have to remember that pin `21` is the CS pin, what do you think about add a constant in the core for it? One possible name is `SPI_FLASH_CS`, similar could be done for the SPI port.

eriknyquist:
Any progress here? can we nudge this one along https://github.com/PaulStoffregen/SerialFlash/pull/35 ?

facchinm:
Merged in mainline, please proceed with this one (I'll push a rebase immadiately :smile: )

eriknyquist:
Thanks @facchinm. Any idea when the next SerialFlash release will be available in the IDE library manager? Or maybe @PaulStoffregen is the one to ask?

eriknyquist:
Also, @facchinm, I think it makes sense for this to be a single commit. Could you squash please?

facchinm:
@eriknyquist three commits make more sense to me but if you prefer squashing it should be a matter of a tick in github's UI during merge :smile:

PaulStoffregen:
> Or maybe @PaulStoffregen is the one to ask?

I have other issues still pending on SerialFlash, mainly support for newer chips and some usability improvements with the examples.  Just yesterday OSH Park delivered a new board I need for testing the largest Micron chips.  Planning to solder it tomorrow and add support for these chips.

If you *really* need a release to push to the boards manager asap, I could do one.  But in just a matter of days I'll be making another release for these new chips.  Would prefer to do just 1 new release.

eriknyquist:
@facchinm fine, as long as each commit makes sense /compiles on its own. @PaulStoffregen also fine, not necessarily in a rush, just wanted the ETA. a few days from now is fine.

kitsunami:
@PaulStoffregen any update on this? Can you review and merge the pull request from @facchinm ?

eriknyquist:
@kitsunami my mistake, Paul has merged this. It was a new release of SerialFlash we were waiting for, so that 101 users can just grab SerialFlash from the Library Manager.

@PaulStoffregen just a friendly ping, do you have any plans for an upcoming release of your SerialFlash library?

PaulStoffregen:
Since November, all my dev cycles have been going into a USB Host library... which has turned out to be the most challenging code I've ever written.  SerialFlash and many others have taken a back seat lately.  I actually have a small pile of new chips to test with SerialFlash.  Hope to get back to this and so much more in another few weeks.

eriknyquist:
Understood @PaulStoffregen, appreciate the update! good luck with the USB host lib :)

sandeepmistry:
@PaulStoffrege another friendly ping, would you be able to tag a new release of your SerialFlash library?

PaulStoffregen:
@sandeepmistry - Done

PaulStoffregen:
Perhaps a bit off-topic... do any of your Intel guys have a contact with Winbond who could get me samples of their new W25M02GVZE chip, for testing with SerialFlash?

SidLeung:
@kitsunami You may have a contact with Winbond?

sandeepmistry:
@PaulStoffregen many thanks!

PaulStoffregen:
Oh, looks like Arrow actually has some in stock.  I just ordered a few.

PR #306, opened 2016-09-29
Remove I2S_MODE_SCK_POL bit from I2S modes

sandeepmistry:
It is incorrectly set to one, table 195 in the data sheet states it should be zero. The non-DMA I2S library also has the bit set to 0.

screen shot 2016-09-29 at 5 34 08 pm


SidLeung:
This bug is corrected in the re-factored I2S library under CurieAudio.

PR #308, opened 2016-09-29
Make g_APinDescription const

eriknyquist:
This has the effect of moving g_APinDescription out of .data (SRAM),
and into .rodata (flash).

Pros:
sketches now have an additional 1160 bytes of SRAM available

Cons:
state of pins is no longer tracked; e.g. when calling digitalWrite,
the full configuration is done each time. See changes in wiring_digital.c
and wiring_analog.c for details. There is inevitably a performance impact
for digitalRead/Write & analogRead/Write, though it is unlikely to be a
noticable one.

@calvinatintel @bigdinotech @SidLeung @russmcinnis @noelpaz please review.

**Do not merge until it is explicitly requested**

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -60,7 +60,7 @@ typedef void (*voidFuncPtr)( void ) ;
 
 /* Types used for the tables below */
 /* TODO - consider using smaller types to optimise storage space */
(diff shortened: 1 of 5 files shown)


noelpaz:
Internal BAT Tests were ran using tT and 101 boards no issues. +1

PR #323, opened 2016-10-19
Expose dccm area

bigdinotech:
-expose dccm area giving the user an extra 8k of memory
-simple implementation of a malloc like function for dccm memory area

### DIFF
--- a/cores/arduino/dccm/dccm_alloc.c
+++ b/cores/arduino/dccm/dccm_alloc.c
@@ -1,0 +1,42 @@
+/*
+Copyright (c) 2016 Intel Corporation.  All right reserved.
+
+This library is free software; you can redistribute it and/or
+modify it under the terms of the GNU Lesser General Public
+License as published by the Free Software Foundation; either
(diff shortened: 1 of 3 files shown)


bigdinotech:

cores/arduino/dccm/dccm_alloc.c
```diff
@@ -0,0 +1,42 @@
...
+
+#include "dccm_alloc.h"
+
+uint16_t dccm_index = 0;
+
+#ifdef __cplusplus
+ extern "C" {
+#endif
+
+uint8_t* dccm_malloc(uint16_t size)
```
eriknyquist: Can you return a `void *`, so that it an be used in the same way as malloc?

i.e. `int *ptr = dccm_malloc(5 * sizeof(int));`

bigdinotech:
@eriknyquist requested changes made

SidLeung:
@bigdinotech Would it be beneficial to define the DCCM segment in flash.ld?  We can define variables that the linker can put into that area.

PR #324, opened 2016-10-20
Move buffers for Wire library to DCCM area

bigdinotech:


bigdinotech:

libraries/Wire/src/Wire.cpp
```diff
@@ -33,7 +33,8 @@ TwoWire::TwoWire(I2C_CONTROLLER _controller_id)
     : rxBufferIndex(0), rxBufferLength(0), init_status(-1),
       controller_id(_controller_id)
 {
-    // Empty
+    rxBuffer = (uint8_t*)dccm_malloc(BUFFER_LENGTH);
+    txBuffer = (uint8_t*)dccm_malloc(BUFFER_LENGTH);
```
eriknyquist: no need to cast, since return type is `void *` the cast will be implicit

eriknyquist: Right you are, we left it in as you can see :)

PR #325, opened 2016-10-20
Allocate RingBuffer buffer into DCCM area

bigdinotech:
-this buffer is used by the uart. Moving it into the DCCM area would
save us 128 bytes of SRAM


bigdinotech:

PR #330, opened 2016-10-21
BLE Central review (Arduino style)

eriknyquist:


libraries/BLE/src/BLEServiceImp.cpp
```diff
@@ -0,0 +1,387 @@
...
+                                                 bledevice);
+    if (NULL == characteristicImp)
+    {
+        return BLE_STATUS_NO_MEMORY;
+    }
+    
+    BLECharacteristicNodePtr node = link_node_create(characteristicImp);
+    if (NULL == node)
+    {
+        delete[] characteristicImp;
```
eriknyquist: This in undefined behaviour-- memory allocated using `new` should bee freed using `delete`, not `delete []`

libraries/BLE/src/BLEProfileManager.cpp
```diff
@@ -0,0 +1,1059 @@
...
+    // Update the service handles
+    while (node != NULL)
+    {
+        serviceCurImp = node->value;
+        if (NULL != serviceCurImp)
+        {
+            servicePrevImp->setEndHandle(serviceCurImp->startHandle() - 1);
+        }
+        
+    pr_debug(LOG_MODULE_BLE, "Curr: start-%d, end-%d", servicePrevImp->startHandle(), servicePrevImp->endHandle());
```
eriknyquist: `pr_debug` call on line 709 will crash the system if `serviceCurImp` is NULL.

libraries/BLE/src/BLECallbacks.cpp
```diff
@@ -0,0 +1,205 @@
...
+    //BLEPeripheralHelper* peripheral = BLECentralRole::instance()->peripheral(conn);// Find peripheral by bt_conn
+    //BLEAttribute* notifyatt = peripheral->attribute(params); // Find attribute by params
+    BLECharacteristicImp* chrc = NULL;
+    BLEDevice bleDevice(bt_conn_get_dst(conn));
+    BLEProfileManager::instance()->characteristic(bleDevice, params->value_handle);
+    
+    //assert(notifyatt->type() == BLETypeCharacteristic);
+    pr_debug(LOG_MODULE_APP, "%s1", __FUNCTION__);
+    
+    chrc->setValue((const unsigned char *)data, length);
```
eriknyquist: It looks to me like `chrc` will *always* be NULL... it is set to NULL on line 98, and then it is not modified again. However you attempt to dereference it here, on line 105, which will lead to a system crash.

libraries/BLE/src/BLECharacteristicImp.cpp
```diff
@@ -0,0 +1,749 @@
...
+    pr_debug(LOG_MODULE_BLE, "%s-%d",__FUNCTION__, __LINE__);
+    if (NULL == descriptorImp)
+    {
+        return BLE_STATUS_NO_MEMORY;
+    }
+    
+    BLEDescriptorNodePtr node = link_node_create(descriptorImp);
+    if (NULL == node)
+    {
+        delete[] descriptorImp;
```
eriknyquist: Undefined behaviour. You cannot use `delete []` to free an object that was allocated with `new`. Use `delete`.

libraries/BLE/src/BLEServiceImp.cpp
```diff
@@ -0,0 +1,387 @@
...
+    pr_debug(LOG_MODULE_BLE, "%s-%d",__FUNCTION__, __LINE__);
+    if (NULL == characteristicImp)
+    {
+        return BLE_STATUS_NO_MEMORY;
+    }
+    
+    BLECharacteristicNodePtr node = link_node_create(characteristicImp);
+    if (NULL == node)
+    {
+        delete[] characteristicImp;
```
eriknyquist: Undefined behaviour, use `delete` not `delete []`

libraries/BLE/src/BLEUtils.cpp
```diff
@@ -0,0 +1,190 @@
...
+        if (mac_str[i] == ':') 
+        {
+            i++;
+            continue;
+        }
+
+        temp[0] = mac_str[i - 1];
+        temp[1] = mac_str[i];
+
+        bd_addr.val[length] = strtoul(temp, NULL, 16);
```
eriknyquist: the array `val` has a size of 6, and it seems that I could cause an overrun (and possibly a crash) by simply passing in a long-enough string (if the string is long enough, then the loop will continue until `length == MAX_UUID_SIZE`, which happens to be 16, not 6).

libraries/BLE/src/BLEDeviceManager.cpp
```diff
@@ -0,0 +1,840 @@
...
+        }
+        else
+        {
+            if (NULL == unused)
+            {
+                unused = temp;
+            }
+        }
+    }
+    pr_debug(LOG_MODULE_BLE, "%s-%d:link_existed-%d unused-%p", __FUNCTION__, __LINE__, link_existed, unused);
```
eriknyquist: line 556 will cause a crash if `unused` is NULL.

libraries/BLE/src/BLEDeviceManager.cpp
```diff
@@ -0,0 +1,840 @@
...
+    
+    if (!link_existed)
+    {
+    pr_debug(LOG_MODULE_BLE, "%s-%d-Device:%s", __FUNCTION__, __LINE__, device.address().c_str());
+        // Send connect request
+        bt_conn_t* conn = bt_conn_create_le(device.bt_le_address(), device.bt_conn_param());
+        if (NULL != conn)
+        {
+            memcpy(unused, device.bt_le_address(), sizeof(bt_addr_le_t));
```
eriknyquist: line 565 will cause a crash if `unused` is NULL.

libraries/BLE/src/BLEProfileManager.cpp
sgbihu: servicePrevImp not be NULL at here. The line 694 can promise this.

libraries/BLE/src/BLEDeviceManager.cpp
sgbihu: Only print the value for debug. Can not make the program crash

libraries/BLE/src/BLEDeviceManager.cpp
sgbihu: Fixed

kitsunami:
Superseded by #337

PR #331, opened 2016-10-21
README.md: remove Serial1 debug information

eriknyquist:
@SidLeung this was added with the big BLE commit (ae78b162), but I'm not sure if we need it in the corelibs README. What do you think?


eriknyquist:
@SidLeung what say you? can we remove this?

PR #333, opened 2016-10-24
Use our own version of copy.exe so that we can copy without cmd.exe

calvinatintel:
in response to https://github.com/01org/corelibs-arduino101/issues/309


calvinatintel:
Need https://github.com/01org/intel-arduino-tools/pull/65 before this is merged. Still testing.

Issue #280, opened 2016-09-02
digitalWrite and analogWrite do not work in combination


khanning:
If you call analogWrite on a PWM capable pin with any value other than 0 and 255, any subsequent  calls to digitalWrite on the same pin are ignored until one of the following events occurs:
1. `pinMode(PIN, OUTPUT)` is called on the pin again
2. You call `analogWrite(PIN, 0)` or `analogWrite(PIN, 255)`

This behavior does not occur on the Arduino UNO. 

Here is a simple sketch that illustrates the issue:

```
void setup() {
  pinMode(9, OUTPUT);
}

void loop() {
  digitalWrite(9, HIGH);
  delay(1000);
  analogWrite(9, 255/2);
  delay(1000);
  digitalWrite(9, LOW);
  delay(1000);
}
```

**On the 101**
Once `analogWrite(9, 255/2)` is called, pin 9 remains at a 50% duty cycle, ignoring all subsequent digitalWrite commands.

**On the UNO**
Pin 9 successfully switches between 100%, 50%, and 0% duty cycle.

eriknyquist:
@bigdinotech can you look into this please?

bigdinotech:
That is actually the expected behavior.
when using analogWrite() the pins are muxed into pwm mode
when using pinMode() the pins are muzxed into gpio mode
We are not muxing during digitalWrite() for performance reasons.
analogWrite(pin, 0) and analogWrite(pin, 255) works because we are actually switching to gpio modes for those values

facchinm:
@bigdinotech I believe that using a bitmap (in RAM) to keep the pin mux state would solve the issue. Since you got much improvement in RAM usage since 1.0.7 we could use some of this spare memory to avoid this kind of problems (which are very time consuming for people porting AVR libraries). For example, we did this on SAM core https://github.com/arduino/Arduino/commit/d1c65ca8fe4fc0a81f6760cc9abcaf140113e214 
Benchmarking the solutions would help also (checking the actual mux should take like a couple of instructions if the array is well crafted)

eriknyquist:
Re-opening since @facchinm's last comment sounds like something we should do

eriknyquist:
Once #355 has been verfied and merged, this issue can be closed.

Issue #334, opened 2016-10-25
analogWrite issue


agdl:
Hi, I found a bug in the analogWrite function. Here the sample sketch I used to reproduce the issue:

```
int led = 3;           // the PWM pin the LED is attached to
int brightness = 0;    // how bright the LED is
int fadeAmount = 5;    // how many points to fade the LED by

// the setup routine runs once when you press reset:
void setup() {
  // declare pin 9 to be an output:
  pinMode(2, INPUT_PULLUP);
  pinMode(led, OUTPUT);
}

// the loop routine runs over and over again forever:
void loop() {
  // set the brightness of pin 9:
  analogWrite(led, brightness);

  // change the brightness for next time through the loop:
  brightness = brightness + fadeAmount;

  // reverse the direction of the fading at the ends of the fade:
  if (brightness <= 0 || brightness >= 255) {
    fadeAmount = -fadeAmount;
  }
  // wait for 30 milliseconds to see the dimming effect
  delay(30);

  if(digitalRead(2) == LOW){
    digitalWrite(led, LOW);
    while(true);
  }
}
```

Basically if you use a pin as PWM and then you use the `digitalWrite(pin, LOW);` the last PWM value stays on the pin.
Obviously writing `analogWrite(pin, 0);` the issue disappears

oqibidipo:
See #280.

kitsunami:
Duplicate to #280

PR #335, opened 2016-10-26
check mux mode when using digitalWrite

bigdinotech:
-check and make sure that the pin is in GPIO_MUX_MODE when doing a
digitalWrite()
-this has a slight performance impact of digitalWrite
-it also comsumes an extra byte in SRAM per pin
-it does however give us the ability to track the muxing state of each
pin

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -26,6 +26,8 @@
 #include <stdbool.h>
 #include <math.h>
 
+#include "pins_arduino.h"
+
 #include "binary.h"
 //#include "itoa.h"
 
@@ -96,6 +98,8 @@ typedef struct _PinDescription
 extern PinDescription g_APinDescription[] ;
 
 extern uint32_t pwmPeriod[4];
+
+extern uint8_t pinmuxMode[NUM_DIGITAL_PINS];
 
 #ifdef __cplusplus
 } // extern "C"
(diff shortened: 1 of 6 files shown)


bigdinotech:
@SidLeung @eriknyquist please review
@khanning @agdl @facchinm this should fix #280 and #334 
as I mentioned before there is a slight performance impact but it seems more negligeable than I initially though

eriknyquist:
I think I2S, I2SDMA and possibly some more need to set pinMuxMode too. See the following two commits, where I removed them, for reference;

https://github.com/01org/corelibs-arduino101/commit/3d3c1f92c2c15724725f71542bc564754ba5d6e3
https://github.com/01org/corelibs-arduino101/commit/1a5e3f7b2d65cf2b81417f85bd24ca455bdaa1fc

agdl:
Works for me!

noelpaz:
@russmcinnis . Can you please verify?

bigdinotech:
@eriknyquist I2S will be added once the @SidLeung finishes the changes for it

russmcinnis:
I test variations of the examples and it works for me too. I was doing some testing using sketch delayMicroseconds() to just measure the frequency of 100 consecutive digitalRead() or digitalWrite() calls without checking any signals. The digitalWrite() function speed with the mux checking added was about 75% of the speed without it.

bigdinotech:
@russmcinnis please test this with CurieSoftwareSerial library.
It should only affect the tx performance of it and the appropriate changes is already in this commit

eriknyquist:
@russmcinnis can you see what SoftwareSerial behaviour is like with these changes?

russmcinnis:
I don't see any difference. I ran the BAT test that hooks up SoftwareSerial pins to Hardware Serial (Serial1) pins. The test runs a lot of transactions and passes. 

Next I ran most of the baud rates by looping:
  mySerial.write(0x1);
I have some tables of the frequency from the scope and I don't see any difference. I thought I would see a difference since before I was measuring digitalWrite(HIGH and LOW) in a tight loop and saw performance fall by 25%.

bigdinotech:
@calvinatintel please merge this one also. Jenkins merger is also failing for this one

PR #340, opened 2016-11-02
ss_spi.c: fix CTRLR0 write in ss_spi_set_data_mode()

eriknyquist:
This fixes two existing issues with the current implementation;
  1. The value to be written (ctrl) is completely overwritten on line 107,
     erasing the previous modifications (perhaps instead of a bitwise-OR?)
  2. On the same line- the 'dataMode' value should be shifted by 4, not by 6
     (see Quark SE C1000 datasheet, Table 229 "Detailed Description of
      Control Register 0 (CTRLR0) for reference)

@bigdinotech @SidLeung @russmcinnis please review & test


russmcinnis:
I just tested setting SPI_MODE0. It isn't hanging, but I haven't verified functionality. I am still seeing the issue with reading all 0s from the SPI FRAM though.

eriknyquist:
Closing-- this fix is invalid now.

PR #342, opened 2016-11-08
BLE KW fixes

eriknyquist:
Temporary pull request so I can keep running KW scans while I'm fixing the issues.
Do not merge.

@noelpaz get the fixes here.

Note that this will NOT be the final API- we have to get it functional before we can tweak how the API looks. Please don't review any of this code yet (or at least, don't comment on style issues).

PR #343, opened 2016-11-10
BLE: combination of 337, 342 & xieqi's commit

eriknyquist:
@SidLeung, test this one please. #342 is still active, just in case this integration breaks something


noelpaz:
Adding @russmcinnis and @nfpaz as reviewers and testers

PR #349, opened 2016-11-29
CurieIMU.cpp: Don't lock interrupts for SPI transfers

eriknyquist:
This is unnecessary- nothing else is contending for access to this
particular SPI controller

@SidLeung @bigdinotech please review

@noelpaz @russmcinnis please test


eriknyquist:
Verified with https://github.com/01org/Intel-Pattern-Matching-Technology/blob/master/examples/DrawingInTheAir/DrawingInTheAir.ino

PR #358, opened 2016-12-20
I2S example sketches failed compliation, Jira 786.

SidLeung:
The latest BLE library consumes more heap space that causes these sketches ran out of heap space for their temporary
buffers.  For now, the size of the tempporary is reduced.  For the near future, there are 3 improvements to address
this issue.  The BLE library will be making use of the DCCM memory instead of heap.  The I2S library buffering space
is going to the DCCM memory too.  The heap space will be increased to maximize the usage of the entire SRAM sapce.


SidLeung:
@bigdinotech  @eriknyquist  Would you please review these changes so that we can make the "pen down" deadline today?

PR #381, opened 2016-12-30
libarc32_arduino101/Makefile: Strip libarc32drv_arduino101.a

eriknyquist:
@calvinatintel @SidLeung please review


eriknyquist:
FYI for testing, the effect of this change is that any code in the .a file is a black box during debugging.
For example, if you set a breakpoint at `Serial.begin()` in your sketch, then keep stepping down until you can go no further, you'll eventually reach some function from the .a file. You won't be able to step into it, or see any source lines inside the function. All you can do is step over it.

SidLeung:
When using JTAG debugger, libarc32drv_arduino101.a is recompiled so that the symbol table will use the path of the local source.  In the Makefile, just comments to tell the user to comment out the stripping for debugging info.   Similar to the comment in platform.txt.  BTW, how much (flash) memory space is saved by stripping the symbol table info in the library?

eriknyquist:
@SidLeung are you sure? can you show me where that is happening?
AFAIK libarc32drv_arduino101.a doesn't get recompiled anywhere unless you run "make" yourself.

eriknyquist:
OK, the final decision for this change is;

Default target for the .a file remains the same. When you run "make", it will build with debug symbols, and not be stripped. This means that CODK users can use "make" to re-build the .a file for debugging purposes.

Also, a new "strip" target has been added to the .a Makefile. Developers should use this target (i.e. "make strip") when re-building the .a file to check into version control.

Repeat- *Use "make strip" when re-building libarc32drv_arduino101.a, if you are going to commit it. Otherwise, paths from the machine you are using will be embedded in the .a file, and will show up in GDB when CODK users try to debug*

SidLeung:
@eriknyquist Thanks for putting in the strip symbol info option in the Makefile.  It looks good.  @calvinatintel Please merge.

PR #395, opened 2017-01-10
CurieIMU.cpp: initialize saved accel/gyro ranges in begin()

eriknyquist:
This ensures that "read[Accelerometer/Gyro]Scaled()" will work as
expected even if a range has not been explicitly set using the API

@russmcinnis to verify


SidLeung:
@eriknyquist Agree with the changes and looks good.  @bigdinotech ?

eriknyquist:
Russ verified that it fixes the issue he saw

PR #396, opened 2017-01-10
Add new library MemoryFree

eriknyquist:
This is the MemoryFree libary from
http://playground.arduino.cc/code/AvailableMemory, rewritten specifically
for Arduino 101 and tinyTILE. It provides some functions for querying
available stack & heap space.


eriknyquist:
Ping @SidLeung @bigdinotech

eriknyquist:
Did you read the README that's in there? can you be more specific about what's missing from it?

PR #401, opened 2017-01-16
Change the link script to optimum the RAM layout

sgbihu:
1. Revert the DCCM change
2. Update the link script


variants/arduino_101/linker_scripts/flash.ld
```diff
@@ -40,8 +40,8 @@ OUTPUT_FORMAT("elf32-littlearc", "elf32-bigarc", "elf32-littlearc")
 MEMORY
     {
     FLASH                 (rx) : ORIGIN = 0x40034000, LENGTH = 152K
-    SRAM                  (wx) : ORIGIN = 0xa800e000, LENGTH = 24K
```
eriknyquist: I thought "rwx" and "wx" were the same thing (i.e. "w" signifies both read and write).  I'm not 100% sure though. Can somebody explain why this is needed?

kitsunami:
This is a performance improvement that will go into the next release.

variants/arduino_101/linker_scripts/flash.ld
eriknyquist: ping.... any answers to my question above?

variants/arduino_101/linker_scripts/flash.ld
```diff
@@ -180,6 +180,14 @@ SECTIONS
 
 	/* Data Closely Coupled Memory (DCCM) */
 /* DCCM Start */
+    .kernelmempool (NOLOAD) :
+    {
+	. = ALIGN(4);
+	__dccm_start = .;
+	*(.kernelmempool)
+	*(".kernelmempool.*")
+	__dccm_end = ALIGN(4);
```
eriknyquist: Note that this does not actually align the location counter-- it *calculates* the next aligned value based on the current location counter, although the counter itself remains unaligned. I think you should do something like this instead;

`. = ALIGN(4);`
`__dccm_end = .;`

eriknyquist (changes requested):

PR #406, opened 2017-01-18
expose all AON-GPIOs

bigdinotech:
-expose AON-GPIOs since they are available on the TinyTILE


bigdinotech:
@eriknyquist @SidLeung Ready for review

PR #407, opened 2017-01-18
Add macro to read value from register

bigdinotech:


bigdinotech:
@eriknyquist @SidLeung Ready for review

system/libarc32_arduino101/common/scss_registers.h
```diff
@@ -38,6 +38,8 @@
 #define MMIO_REG_ADDR(addr) ((volatile uint32_t *)(addr))
 #define MMIO_REG_VAL_FROM_BASE(base, offset) \
     (*((volatile uint32_t *)((base)+(offset))))
+    
+#define REG_VAL(addr) (*(uint32_t *)(addr))
```
eriknyquist: Why do we need this? we already have MMIO_REG_VAL. Line 37.

bigdinotech: It was not necessary to use volatile for all my register accesses in the Power Management library. I could just make everything volatile which should also work.

eriknyquist: No need to make anything volatile-- MMIO_REG_VAL casts the passed pointer as a volatile, which forces volatile semantics on whatever that pointer happens to point to.

And, as was reinforced last week during our course-- you always use volatile to access memory mapped HW! So always using MMIO_REG_VAL is a good idea.

PR #417, opened 2017-01-30
Use single dataReady() method, instead of accelDataReady/gyroDataReady

eriknyquist:
This method is a bit more flexible. There is a version with no
parameters, which will return true if all enabled sensors have new
data. There is one overload, that takes a single 'flags' parameter,
which returns true if all enabled *and* selected sensors have new data.
Sensors are selectable through pre-defined bit flags. For example;

    CurieIMU.dataReady() : this will return true if both the accel and gyro
    are enabled, and both have new data ready.

    CurieIMU.dataReady(ACCEL | GYRO) : this will return true if both the
    accel and gyro have new data ready (if one of the passed sensors is not
    enabled, it is ignored and the function will behave as if that sensor was not
    specified)

    CurieIMU.dataReady(GYRO) : this will return true if the gyro is
    enabled and has new data ready.

    CurieIMU.dataReady(ACCEL) : this will return true if the accel is
    enabled and has new data ready.

This also involved changing the 'begin' method to allow selective
enabling of sensors, e.g.

    CurieIMU.begin() : same as it always has, initializes IMU, enabling
    both accel and gyro

    CurieIMU.begin(ACCEL | GYRO) : initializes IMU, enabling both accel
    and gyro

    CurieIMU.begin(ACCEL) : initializes IMU, enabling accel only

    CurieIMU.begin(GYRO) : initializes IMU, enabling  gyro only

### DIFF
--- a/libraries/CurieIMU/keywords.txt
+++ b/libraries/CurieIMU/keywords.txt
@@ -14,6 +14,7 @@ CurieIMUClass	KEYWORD1
 
 begin	KEYWORD1
 
+dataReady   KEYWORD1
 getGyroRate	KEYWORD1
 setGyroRate	KEYWORD1
 getAccelerometerRate	KEYWORD1
@@ -83,6 +84,9 @@ CurieIMU	KEYWORD2
 #######################################
 # Constants (LITERAL1)
 #######################################
+
+ACCEL   LITERAL1
+GYRO    LITERAL1
 
 X_AXIS	LITERAL1
 Y_AXIS	LITERAL1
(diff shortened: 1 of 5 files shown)


eriknyquist:
@russmcinnis to verify

PaulStoffregen:
Just curious, why dataReady?

In the Arduino world, isn't the well established convention to name this "available"?

eriknyquist:
`dataReady()` because it does what it says on the tin..... `available` generally tells you how many buffered incoming items are available. `dataReady()` just tells you whether the next un-read sample is available from a particular sensor, no buffering. True or false.

eriknyquist:
@PaulStoffregen also, @facchinm and @tigoe and I discussed this back and forth for a bit, and this API is the one we all agreed on :)

PR #418, opened 2017-01-31
Jira 795, read() is blocking selectable, git 383

SidLeung:
Mods:
1. Add an input parameter to BLECharacteristic::read() for blocking
   call selection.
2. By default, read() is a blocking call.

File changes:
1. libraries/CurieBLE/src/BLECharacteristic.cpp:
   - Method read() added blocking selection.
2. libraries/CurieBLE/src/BLECharacteristic.h:
   - Prototype for read(), default is blocking.
3. libraries/CurieBLE/src/internal/BLECallbacks.cpp:
   - Added parameter checking in profile_read_rsp_process().
4. libraries/CurieBLE/src/internal/BLECharacteristicImp.cpp
   - The implementation of read() added waiting for resp if
     blocking is selected.
5. libraries/CurieBLE/src/internal/BLECharacteristicImp.h:
   - prototyping.

### DIFF
--- a/libraries/CurieBLE/src/internal/BLECharacteristicImp.h
+++ b/libraries/CurieBLE/src/internal/BLECharacteristicImp.h
@@ -146,18 +146,18 @@ public:
     /**
      * @brief   Schedule the read request to read the characteristic in peripheral
      *
-     * @param[in]   none
+     * @param[in]   blocked    Flag the call is blocked or un-blocked
      *
      * @return  bool    Indicate the success or error
      *
-     * @note  Only for central device
+     * @note  Only for GATT client
+     *        Default it is block call as per Arduino request
      */
-    bool read();
+    bool read(bool blocked = true);
     
     /**
      * @brief   Schedule the write request to update the characteristic in peripheral
      *
-     * @param[in]   peripheral   The peripheral device that want to be updated
      * @param[in]   value       New value to set, as a byte array.  Data is stored in internal copy.
      * @param[in]   length      Length, in bytes, of valid data in the array to write.
      *                      Must not exceed maxLength set for this characteristic.
@@ -328,7 +328,7 @@ private:
     bt_gatt_subscribe_params_t _sub_params;
     bool        _subscribed;
     
-    bool _reading;
+    volatile bool _reading;
     static volatile bool _gattc_writing;
     bt_gatt_read_params_t _read_params; // GATT read parameter
     
--- a/libraries/CurieBLE/src/BLECharacteristic.cpp
+++ b/libraries/CurieBLE/src/BLECharacteristic.cpp
@@ -367,14 +367,14 @@ bool BLECharacteristic::canUnsubscribe()
     return retVar;
 }
 
-bool BLECharacteristic::read()
+bool BLECharacteristic::read(bool blocked)
 {
     bool retVar = false;
     BLECharacteristicImp *characteristicImp = getImplementation();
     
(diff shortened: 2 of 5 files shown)


SidLeung:
@bigdinotech @eriknyquist , please review this code change.  @noelpaz @russmcinnis , please verify this fix.  Please use the example sketch, peripheral_explorer, as a guideline for testing the resolution.

libraries/CurieBLE/src/internal/BLECharacteristicImp.h
```diff
@@ -146,18 +146,18 @@ class BLECharacteristicImp: public BLEAttribute{
...
-     * @param[in]   none
+     * @param[in]   blocked    Flag the call is blocked or un-blocked
      *
      * @return  bool    Indicate the success or error
      *
-     * @note  Only for central device
+     * @note  Only for GATT client
+     *        Default it is block call as per Arduino request
      */
-    bool read();
```
eriknyquist: So this doesn't change any existing code, right? i.e. if I call `read()` with no arguments, it will block by default, yes?

libraries/CurieBLE/src/internal/BLECharacteristicImp.h
```diff
@@ -324,7 +324,7 @@ class BLECharacteristicImp: public BLEAttribute{
     bt_gatt_subscribe_params_t _sub_params;
     bool        _subscribed;
     
-    bool _reading;
+    volatile bool _reading;
```
eriknyquist: Why does this need to be volatile?

SidLeung:
@noelpaz , @russmcinnis , please provide feed back on this PR.

libraries/CurieBLE/src/internal/BLECharacteristicImp.h
sgbihu: You are correct.

libraries/CurieBLE/src/internal/BLECharacteristicImp.h
sgbihu: Please note last two changes in BLECharacteristicImp.cpp. It need to make compiler not optimumize the code.

eriknyquist: OK, so *why* do you need to force optimization off? Who is accessing this variable out of normal execution flow? Describe the situation that will occur if `_reading` is not `volatile`.

sgbihu: When read response come back, the setValue will be called and clear the reading flag. Because the response is asynchronous. So the below code may optimized for the value was set true before call while.

_reading = true;

while(_reading)
{
......
}

noelpaz:
@russmcinnis  is testing this functionality.  But we cannot do an end to end testing unless we resolve the conflicts. Pinging @SidLeung

SidLeung:
@noelpaz , @russmcinnis , conflict resolved, please proceed with system testing.  Thanks.

SidLeung:
@noelpaz @russmcinnis , please retest this PR.  It has resolved the hanging issue due to disconnection in the middle of reading.

@bigdinotech @eriknyquist , please review the additional code change for disconnection detection.

SidLeung:
@yashaswini-hanji , please merge the PR to main and close out Jira 795.

yashaswini-hanji:
PR merged and Jira 795 is closed

PR #419, opened 2017-01-31
Fix arduino101load argument passing if binary path contains spaces

facchinm:
This issue was first reported to the support channel and was solved by the proposed patch

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -103,7 +103,7 @@ tools.arduino101load.cmd.path={runtime.tools.arduino101load.path}/arduino101load
 tools.arduino101load.upload.params.verbose=verbose
 tools.arduino101load.upload.params.quiet=quiet
 
-tools.arduino101load.upload.pattern="{cmd.path}" "{runtime.tools.arduino101load.path}/x86/bin" {build.path}/{build.project_name}.bin {serial.port} "{upload.verbose}" {ble.fw.string} {ble.fw.position}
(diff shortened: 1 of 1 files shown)


SidLeung:
@calvinatintel @bigdinotech @eriknyquist , please review the change.  @noelpaz @russmcinnis , please perform verification testing.  Will merge the change to main trunk upon completion.  Targeted for Deneb.

calvinatintel:
This is a fine solution, but I think `{build.path}` should be provided with the spaces escaped from the upstream.

facchinm:
http://forum.arduino.cc/index.php?topic=448542.0 shows the same problem indeed

calvinatintel:
@russmcinnis @noelpaz Please test

novashah:
@russmcinnis @noelpaz Is this tested?

noelpaz:
I tested this by saving the skecth in a folder with spaces in the name pleace note than when I named my Sketch "S P A C E D  Blink", it renamed it into S_P_A_C_E_S__Blink  and the IDE spit out this messgae:

The sketch name had to be modified. Sketch names can only consist
of ASCII characters and numbers (but cannot start with a number).
They should also be less than 64 characters long.

@calvinatintel explained that it is to verify that the enclosing folders that will have spaces. Fix is verified

noelpaz:
@facchinm  -- fix verified. @yashaswini-hanji  -- Please Squash and Merge

yashaswini-hanji:
PR merged

PR #421, opened 2017-02-01
freeMemory.ino: add introductory comment

eriknyquist:


eriknyquist:
Also note that this is basically just a port of http://playground.arduino.cc/Code/AvailableMemory

Which is linked in the README too.

PR #429, opened 2017-02-08
Fix the UART change makes BLE library failed

sgbihu:
1. The uart_fifo_fill doesn't check the FIFO's length and failed to send out all data
2. Check the FIFO status register before write data in FIFO

### DIFF
--- a/cores/arduino/UARTClass.cpp
+++ b/cores/arduino/UARTClass.cpp
@@ -168,7 +168,7 @@ int UARTClass::read( void )
 
 void UARTClass::flush( void )
 {
-  while (_tx_buffer->_iHead != _tx_buffer->_iTail); //wait for transmit data to be sent
+  while (_tx_buffer->_iHead != (_tx_buffer->_iTail)); //wait for transmit data to be sent
   // Wait for transmission to complete
   while(!uart_tx_complete(CONFIG_UART_CONSOLE_INDEX));
 }
@@ -183,8 +183,7 @@ size_t UARTClass::write( const uint8_t uc_data )
   {
     // If busy we buffer
     int l = (_tx_buffer->_iHead + 1) % UART_BUFFER_SIZE;
-    while (_tx_buffer->_iTail == l)
-      ; // Spin locks if we're about to overwrite the buffer. This continues once the data is sent
+    while (_tx_buffer->_iTail == l); // Spin locks if we're about to overwrite the buffer. This continues once the data is sent
 
     _tx_buffer->_aucBuffer[_tx_buffer->_iHead] = uc_data;
     _tx_buffer->_iHead = l;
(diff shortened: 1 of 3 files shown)


cores/arduino/UARTClass.cpp
```diff
@@ -183,8 +183,7 @@ size_t UARTClass::write( const uint8_t uc_data )
   {
     // If busy we buffer
     int l = (_tx_buffer->_iHead + 1) % UART_BUFFER_SIZE;
-    while (_tx_buffer->_iTail == l)
-      ; // Spin locks if we're about to overwrite the buffer. This continues once the data is sent
+    while (*(volatile int*)&(_tx_buffer->_iTail) == l); // Spin locks if we're about to overwrite the buffer. This continues once the data is sent
```
eriknyquist: It looks like the RingBuffer class is only used by the UART driver, so I think we can just declare _iTail as volatile in RingBuffer.h, rather than forcing volatile semantics for each access.

noelpaz:
We need a unit test  to be able to test this. How was this bug found?

cores/arduino/UARTClass.cpp
sgbihu: I only changed the logic. So this change is previous version that I revert. And this need change the ring buffer. Do you still think we need change it? If you think this is necessary, I can change it.

SidLeung: The fact of the matter is that any variable that can be changed by one task (eg. foreground) and being polled by another (eg. background) needs the actual access to the variable itself, no staging of the variable allowed.  Thus, no compiler optimization can be used when accessing this kind of variable and that means it can only be volatile variable.  This is to guarantee that any code will have access to the real value of this kind of variable without being aware of the fact that the variable can be changed in real time.

I concur that the head, tail, and overflow variable in RingBuffer class to be made volatile.

eriknyquist: Agree with @SidLeung. RingBuffer::store_char() is ccalled in the ISR, and used all 3 variables. Let's declare them all as volatile.

sgbihu:
@noelpaz this is updated UART code make BLE failed issue. Previously, we just revert it. Now I try to fix the issue based on original commit.

SidLeung:
@noelpaz , not ready for testing yet.  Will ping you on that.  The testing of this fix is quite simple, no new sketches needed.  The original PR caused the UART interface to the Nordic chip to fail.  This PR added the correction to make the original PR working.  You can test it with any BLE sketches, Central and Peripheral.

@sgbihu , please proceed with the volatile changes unless there are concerns.

eriknyquist:
@sgbihu please squash into one commit

SidLeung:
Please note this PR is replace by,

https://github.com/01org/corelibs-arduino101/pull/436

sgbihu:
Done

SidLeung:
This PR is replaced by #436, closing this one here.

PR #431, opened 2017-02-08
Jira 804 BLE characteristic buffer allocation issue, git 366

sgbihu:
…ue # 366

1. Update the length when initial by characteristic


SidLeung:
This PR is replaced by #433  and it is closed.

Issue #420, opened 2017-02-01
CurieBLE: add broadcast property to characteristic


don:
I was porting some [BLEPeripheral](https://github.com/sandeepmistry/arduino-BLEPeripheral) code to the Arduino 101 and noticed that BLEBroadcast is missing from CurieBLE.

<img width="1337" alt="screen shot 2017-01-30 at 11 14 56 pm" src="https://cloud.githubusercontent.com/assets/3133/22491526/2848988e-e7f2-11e6-9275-2383019de632.png">

I'd like to be able to add the [BLEBroadcast property](https://github.com/sandeepmistry/arduino-BLEPeripheral/blob/master/API.md#constructor-2) to a Characteristic. [setValue](https://github.com/sandeepmistry/arduino-BLEPeripheral/blob/master/API.md#set-value) should update the advertising data when the characteristic's value changes.

SidLeung:
Hi, Don:

I have not use the broadcast() API before in the previous library.  However, I have used the BLENotify option in a Characteristic in one of my Peripheral sketches.  To notify the connected Central for data ready, my sketch just need to use the setValue() API to set the local buffer with the data.  The BLE library takes care of the notification process.  There is no specific API to call to prompt the library to perform a notification.  I would think the broadcast option to be similar.

don:
@SidLeung BLENotify requires a connection between the client and the server.

BLEBroadcast allows me to send data without a connection. I created a [demo sketch](https://gist.github.com/don/536d9f1d605d9bfffcd99b5716a72cb4) for you. I ran it on a RFduino, but it'll work on RedBear Nano or other boards supported by [BLEPeripheral](https://github.com/sandeepmistry/arduino-BLEPeripheral#compatible-hardware). The peripheral increments a counter and sets the characteristic value. Since the characteristic is marked as broadcast, the advertising data is updated whenever the value changes. This allows a my phone to get the counter value when scanning. I don't need to connect to get the data.

    // Broadcast Characteristic Value
    // Simple counter that broadcasts a value
    #include <BLEPeripheral.h>

    uint8_t value = 0;

    BLEPeripheral peripheral;
    BLEService service = BLEService("EEE0");
    BLEShortCharacteristic characteristic = BLEShortCharacteristic("EEE1", BLERead | BLENotify | BLEBroadcast);

    void setup() {
      Serial.begin(9600);

      peripheral.setLocalName("BLEBroadcast");
[...]

PR #437, opened 2017-02-13
Jira 832 Sample Sketch to exercise the BLEbroadcast option.

sgbihu:
1. Add example sketch
    libraries/CurieBLE/examples/peripheral/broadcast

### DIFF
--- a/libraries/CurieBLE/examples/peripheral/broadcast/broadcast.ino
+++ b/libraries/CurieBLE/examples/peripheral/broadcast/broadcast.ino
@@ -1,0 +1,30 @@
+#include <CurieBLE.h>
+
+uint8_t value = 0;
+
+BLEService service = BLEService("EEE0");
+BLEShortCharacteristic characteristic = BLEShortCharacteristic("EEE1", BLERead | BLENotify | BLEBroadcast);
+
+void setup() {
+  Serial.begin(9600);
+
+  BLE.setLocalName("BLEBroadcast");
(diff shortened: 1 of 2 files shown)


SidLeung:
@sgbihu , I think you are using the sketch from Don Coleman @ gitHub.  I am asking him to put in a licensing header before we pull it in as part of the CoreLibs release.

@calvinatintel , @yashaswini-hanji , is there a procedure to take sketch from GItHub and use it in our coreLibs?  I suppose we just need the proper licensing.  Am I correct?

calvinatintel:
What's the license of the original sketch? Can you give me a link to it?

SidLeung:
@calvinatintel , @yashaswini-hanji ,  it is code from someone's gitHub request,

https://github.com/01org/corelibs-arduino101/issues/420

I did ask him for licensing.  Please check out the end of the threat.

SidLeung:
In light of the fact that the sample sketch is not public, another PR without the sample sketch is generated,

https://github.com/01org/corelibs-arduino101/pull/438

This PR will be closed.

SidLeung:
This PR is replaced by #438.

noelpaz:
I tested the sketch and the way to test this in NRF is to just scan but you cannot connect and do not connect. You should see the Service data incrementing.

The caveat is that if you central app stops scanning then the Service data will stop. Just refresh the scan and you should see the incrementing value.

Issue #384, opened 2016-12-30
CurieBLE: Characteristic UUID's have incorrect value of 0000


sandeepmistry:
When using the [pizza example from bleno](https://github.com/sandeepmistry/bleno/tree/master/examples/pizza) on my Mac when the peripheral explorer CurieBLE example, a few characteristics have UUID values of `0000` which is incorrect. Here's the output from the Serial Monitor:

```
BLE Central - Peripheral Explorer
Found C8:69:CD:41:84:A2 '' 
Found E4:1E:14:D2:31:E7 '' 
Found AC:BC:32:7C:FA:26 '' 13333333-3333-3333-3333-333333333337
Connecting ...
Connected
Discovering attributes ...
Attributes discovered

Device name: 
Service 1800
Service 1801
Service 180a
	Characteristic 2a29, properties 0x2, value 0x4170706C6520496E63
	Characteristic 2a24, properties 0x2, value 0x4D6163426F6F6B50726F31322C31
Service d0611e78-bbb4-4591-a5f8-487910ae4366
	Characteristic 0000, properties 0x24
Service 13333333-3333-3333-3333-333333333337
	Characteristic 0000, properties 0x10, value 0x00
	Characteristic 0000, properties 0x10, value 0x0000
	Characteristic 0000, properties 0x24

Disconnecting ...
Disconnected
```

If I connect to the Mac peripheral using LightBlue on iOS the characteristic UUID's value match what I expect.

sgbihu:
Can you describe your test instructions? I suspect the characteristic's UUID are not sub-UUID with services. And that makes the stack return an error. The characteristic's UUID in the below bold text is different with service UUID, then the discover can success. If the other is not same. I think it may have the error you described.

19b1**0000**-e8f2-537e-4f6c-d104768a1214

sandeepmistry:
I think this is specific to bleno on OS X, as OS X has some "secure" characteristics used for hand off with iOS devices.

Test instructions:
1. Obtain a Bluetooth 4.0 capable Mac (I'm using 10.11.x)
2. Install Node.js
3. Make a test directory
4. Run `npm install bleno`
5. `cd node_modules/bleno`
6. Run `node examples/pizza/peripheral.js`

sgbihu:
Hi @sandeepmistry ,

The 0000 UUID is expected after I review your test code. Because the UUID not satisfy the Nordic's stack. It will return unknown UUID and this UUID is 0.
[...]

PR #440, opened 2017-02-15
Jira793 Central returned incorrect Characteristic value for some Serv…

sgbihu:
…ice Git #384

1. The root cause is Nordic stack can't return long UUID when discover
    - Change the discover process.
    - Add read operation after discover returned invalid UUID.


SidLeung:
This PR is replaced by #451

PR #447, opened 2017-02-20
Jira 857, 858, BLEPeripherial connect issues

### LINKED ISSUE #444: CurieBLE: BLEPeripheral::connected() always returns false
Observed while debugging: https://github.com/01org/corelibs-arduino101/issues/377#issuecomment-280759093

```arduino
#include <CurieBLE.h>

BLEPeripheral blePeripheral;

void setup() {
  Serial.begin(9600);
  while(!Serial);

  blePeripheral.setLocalName("test");

  blePeripheral.begin();
}

void loop() {
  if (blePeripheral.connected()) {
    Serial.println("connected");

    while (blePeripheral.connected());

    Serial.println("disconnected");
  }

}
```

if i connect from a central, "connected" is never printed.

sgbihu:
Jira 857. BLEPeripheral::connect() return false, git 444.
Jira 858. Cannot trigger BLEConnected and BLEDisconnected events, git 445.

Both related connection issues are address in this PR.

### DIFF
--- a/libraries/CurieBLE/src/BLEPeripheral.cpp
+++ b/libraries/CurieBLE/src/BLEPeripheral.cpp
@@ -21,7 +21,7 @@
 
 #include "BLEPeripheral.h"
 
-static BLEPeripheralEventHandler m_eventHandlers[BLEDeviceLastEvent];
+static BLEPeripheralEventHandler m_eventHandlers[BLEDeviceLastEvent] = {NULL, NULL, NULL, NULL};
 
 void bleBackCompatiblePeripheralConnectHandler(BLEDevice central) 
 {
@@ -154,7 +154,6 @@ void BLEPeripheral::init()
     if (!_initCalled)
     {
         BLE.begin();
-        memset(m_eventHandlers, 0, sizeof(m_eventHandlers));
         _initCalled = true;
     }
 }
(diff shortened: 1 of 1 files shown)


SidLeung:
@bigdinotech , @eriknyquist , please review the code changes.

@noelpaz , @russmcinnis , please system test the changes.  Please use the sample sketch in the gitHub issues to check out the fixes.

libraries/CurieBLE/src/BLEPeripheral.cpp
```diff
@@ -21,7 +21,7 @@
 
 #include "BLEPeripheral.h"
 
-static BLEPeripheralEventHandler m_eventHandlers[BLEDeviceLastEvent];
+static BLEPeripheralEventHandler m_eventHandlers[BLEDeviceLastEvent] = {NULL, NULL, NULL, NULL};
```
eriknyquist: This is unncessary-- `m_evenHandlers` is declared static, so it will be located in .bss
All elements will be initialised to 0 by default.

SidLeung: Agree.  The global variable is in _bss segment and it is clear out on boot.  In addition, the initialization is hard code to 4 which may not be true in future.

@sgbihu , please remove the initialization, stay with the previous code for this item.  Please make the correction and make another push ASAP.  Pencil down this Friday.

SidLeung: @eriknyquist , correction made and it is in a new PR, #455.

SidLeung:
This PR is replaced by two separated PRs:  #454 and #455.

PR #451, opened 2017-02-22
Jira 793 Support Characteristic UUID that is not subset of Service, g…

SidLeung:
…it 384

Root cause:
  - Nordic stack does not return long UUID for Characteristic
    discovery.  It is expecting Characteristic UUID is a
    subset of the Service UUID (eg. a sequential increament of
    the Service UUID).
  - In order for the Central mode (to behave like Apple) to
    discover long Characteristic UUID, the process is now done
    in two steps (similar to the Service discovery).  First,
    initiate a read of the Characteristic and then discover the
    long UUID upon successful read.

Code Mods:

1. BLECallbacks.cpp:
   - Added a call back to handle the read Characteristic
     completion event.
2. BLECallbacks.h:
   - Prototyping.
3. BLEProfileManager.cpp:
   - Added getServiceBySubHandle() to enable the call back to
     locate the Service of a discovered Characteristic.
4. BLEProfileManager.h:
   - Prototyping.
5. BLEServiceImp.cpp:
   - This is where the two steps Characteristic discovery is
     implemented.  At discoverAttributes(), reading of
     Characteristic is first initiated.  When the stack
     completes the operation, extract the UUID thereafter.
   - To support this two steps process, additional functions,
     discoverNextCharacteristic and readCharacteristic, are
     added.
6. BLEServiceImp.h:
   - Prototyping.

    - Change the discover process.
    - Add read operation after discover returned invalid UUID.

### DIFF
--- a/libraries/CurieBLE/src/internal/BLECallbacks.cpp
+++ b/libraries/CurieBLE/src/internal/BLECallbacks.cpp
@@ -172,6 +172,23 @@ uint8_t profile_service_read_rsp_process(bt_conn_t *conn,
     return ret;
 }
 
+uint8_t profile_characteristic_read_rsp_process(bt_conn_t *conn, 
+                                                 int err,
+                                                 bt_gatt_read_params_t *params,
+                                                 const void *data, 
+                                                 uint16_t length)
+{
+    BLEDevice bleDevice(bt_conn_get_dst(conn));
+    BLEServiceImp* service_imp = BLEProfileManager::instance()->getServiceBySubHandle(bleDevice, params->single.handle);
+    
+    uint8_t ret = service_imp->characteristicReadRspProc(conn, 
+                                                         err,
+                                                         params,
+                                                         data,
+                                                         length);
+    pr_debug(LOG_MODULE_BLE, "%s-%d:ret-%d", __FUNCTION__, __LINE__, ret);
+    return ret;
+}
 
 
 void bleConnectEventHandler(bt_conn_t *conn, 
--- a/libraries/CurieBLE/src/internal/BLEProfileManager.cpp
+++ b/libraries/CurieBLE/src/internal/BLEProfileManager.cpp
@@ -507,6 +507,40 @@ BLEServiceImp* BLEProfileManager::service(const BLEDevice &bledevice, int index)
     return serviceImp;
 }
 
+BLEServiceImp* BLEProfileManager::getServiceBySubHandle(const BLEDevice &bledevice, uint16_t handle) const
+{
+    BLEServiceImp* serviceImp = NULL;
+    uint16_t start_handle;
+    uint16_t end_handle;
+
+    const BLEServiceLinkNodeHeader* serviceHeader = getServiceHeader(bledevice);
+    if (NULL == serviceHeader)
+    {
+        // Doesn't find the service
+        return NULL;
+    }
+    BLEServiceNodePtr node = serviceHeader->next;
+    
+    while (node != NULL)
+    {
+        serviceImp = node->value;
+        start_handle = serviceImp->startHandle();
+        end_handle = serviceImp->endHandle();
+        if (handle >= start_handle && handle <= end_handle)
+        {
+            break;
+        }
+        node = node->next;
+    }
+    
+    if (NULL == node)
+    {
+        serviceImp = NULL;
+    }
+
+    return serviceImp;
+}
+
 void BLEProfileManager::handleConnectedEvent(const bt_addr_le_t* deviceAddr)
 {
     int index = getUnusedIndex();
--- a/libraries/CurieBLE/src/internal/BLEServiceImp.cpp
+++ b/libraries/CurieBLE/src/internal/BLEServiceImp.cpp
@@ -26,6 +26,7 @@
 #include "BLECharacteristicImp.h"
 
 bt_uuid_16_t BLEServiceImp::_gatt_primary_uuid = {BT_UUID_TYPE_16, BT_UUID_GATT_PRIMARY_VAL};
+bt_gatt_read_params_t BLEServiceImp::_read_params;
 
 bt_uuid_t *BLEServiceImp::getPrimayUuid(void)
 {
@@ -36,6 +37,7 @@ BLEServiceImp::BLEServiceImp(BLEService& service):
     BLEAttribute(service.uuid(), BLETypeService), 
     _start_handle(0),
     _end_handle(0xFFFF),
+    _reading(false),
     _cur_discover_chrc(NULL)
 {
     memset(&_characteristics_header, 0, sizeof(_characteristics_header));
@@ -46,6 +48,7 @@ BLEServiceImp::BLEServiceImp(const bt_uuid_t* uuid):
     BLEAttribute(uuid, BLETypeService),
     _start_handle(0),
     _end_handle(0xFFFF),
+    _reading(false),
     _cur_discover_chrc(NULL)
 {
     memset(&_characteristics_header, 0, sizeof(_characteristics_header));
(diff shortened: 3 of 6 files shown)


SidLeung:

@russmcinnis @noelpaz , this change is to make our library to behave more like an Apple Central.  The capability of discovering Characteristic UUID that is not sequential increment of the Service UUID.  Please use an existing Peripheral sketch, modify the Characteristic UUID so that it is different from its Service UUID, and use a Central sketch to discover the Characteristic.

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
```diff
@@ -347,44 +370,143 @@ uint8_t BLEServiceImp::discoverResponseProc(bt_conn_t *conn,
...
+    return retVal;
+}
+
+void BLEServiceImp::discoverNextCharacteristic(BLEDevice &bledevice)
+{
+    const BLECharacteristicLinkNodeHeader* chrcHeader = &_characteristics_header;
+    BLECharacteristicImp* chrcCurImp = NULL;
+    BLECharacteristicNodePtr node = chrcHeader->next;
+    
+    //pr_debug(LOG_MODULE_BLE, "%s-%d: node-%p",__FUNCTION__, __LINE__, node);
```
eriknyquist: commented pr_debug? I forgot what our policy is on this. Personally I'd take it out.

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
```diff
@@ -347,44 +370,143 @@ uint8_t BLEServiceImp::discoverResponseProc(bt_conn_t *conn,
...
+    // Discover next service
+    while (node != NULL)
+    {
+        chrcCurImp = node->value;
+        
+        if (NULL == _cur_discover_chrc)
+        {
+            bool result = chrcCurImp->discoverAttributes(&bledevice);
+            pr_debug(LOG_MODULE_BLE, "%s-%d",__FUNCTION__, __LINE__);
```
eriknyquist: pr_debug? should this line be removed?

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
```diff
@@ -347,44 +370,143 @@ uint8_t BLEServiceImp::discoverResponseProc(bt_conn_t *conn,
...
-                pr_debug(LOG_MODULE_BLE, "%s-%d",__FUNCTION__, __LINE__);
-                if (result == true)
-                {
-                    // Record the current discovering service
-                    _cur_discover_chrc = chrcCurImp;
-                    break;
-                }
+                // Record the current discovering service
+                _cur_discover_chrc = chrcCurImp;
+                break;
```
eriknyquist: Use `return` here instead of `break`. The intention is clearer, and it is easier to follow the loop logic.

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
```diff
@@ -347,44 +370,143 @@ uint8_t BLEServiceImp::discoverResponseProc(bt_conn_t *conn,
...
+        return false;
+    }
+    // Send read request
+    retval = bt_gatt_read(conn, &_read_params);
+    bt_conn_unref(conn);
+    if (0 == retval)
+    {
+        _reading = true;
+    }
+    pr_debug(LOG_MODULE_BLE, "%s-%d", __FUNCTION__, __LINE__);
```
eriknyquist: I believe all these calls to pr_debug will incur some serial I/O. If this is the case, please remove them..

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
```diff
@@ -347,44 +370,143 @@ uint8_t BLEServiceImp::discoverResponseProc(bt_conn_t *conn,
...
-                    _cur_discover_chrc = chrcCurImp;
-                    break;
-                }
+                // Record the current discovering service
+                _cur_discover_chrc = chrcCurImp;
+                break;
             }
-            else if (_cur_discover_chrc == chrcCurImp)
+        }
+        else if (_cur_discover_chrc == chrcCurImp)
```
eriknyquist: It looks to me like perhaps just `else` would be sufficient here. @sgbihu could you confirm/deny please?

russmcinnis:
Tested this with and without pull request 451. I used these in the led peripheral. I was able to prove it breaks without PR 451 and passes with PR 451. Also ran the system tests and this holds up.

// create service
BLEService               ledService("18c2444459e3648f5e7dc215879b0000");

// create switch characteristic
BLECharCharacteristic    switchCharacteristic("19b10001e8f2537e4f6cd104768a1214", BLERead | BLEWrite);

libraries/CurieBLE/src/internal/BLECallbacks.cpp
```diff
@@ -172,6 +172,23 @@ uint8_t profile_service_read_rsp_process(bt_conn_t *conn,
...
+{
+    BLEDevice bleDevice(bt_conn_get_dst(conn));
+    BLEServiceImp* service_imp = BLEProfileManager::instance()->getServiceBySubHandle(bleDevice, params->single.handle);
+    
+    uint8_t ret = service_imp->characteristicReadRspProc(conn, 
+                                                         err,
+                                                         params,
+                                                         data,
+                                                         length);
+    pr_debug(LOG_MODULE_BLE, "%s-%d:ret-%d", __FUNCTION__, __LINE__, ret);
```
sgbihu: This is only for debug. For the release, the log will never output to the terminal.

libraries/CurieBLE/src/internal/BLEProfileManager.cpp
```diff
@@ -507,6 +507,40 @@ BLEServiceImp* BLEProfileManager::service(const BLEDevice &bledevice, int index)
...
+
+    const BLEServiceLinkNodeHeader* serviceHeader = getServiceHeader(bledevice);
+    if (NULL == serviceHeader)
+    {
+        // Doesn't find the service
+        return NULL;
+    }
+    BLEServiceNodePtr node = serviceHeader->next;
+    
+    while (node != NULL)
```
sgbihu: I like use break instead of return. Because we may extend the code in the feature, we don't need add the code in the loop if we use the break. Just add the code at the end of the function. But if you persist in your style, I can update the code.

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
sgbihu: Deleted.

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
sgbihu: Same as before.

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
sgbihu: Not correct. The else-if is necessary, because the code need find the previous characteristic, and find next one by previous characteristic. So this is try to reset it to NULL.

libraries/CurieBLE/src/internal/BLEServiceImp.cpp
sgbihu: If we disable the debug, the serial I/O will not be called.

libraries/CurieBLE/src/internal/BLEProfileManager.cpp
eriknyquist: no, it's OK. I understand your explanation, just looking for the reasoning :)

libraries/CurieBLE/src/internal/BLECallbacks.cpp
eriknyquist: Understood. I tested it out also and saw no serial output.

SidLeung:
Code review approach, system testing passed, proceed with merging to main trunk.

@yashaswini-hanji , please merge this PR and close out the Jira ticket.

yashaswini-hanji:
PR merged

Issue #456, opened 2017-02-23
CurieBLE: Large characteristics (64 bytes) crash board when attempting to write to them


kgillespieanaren:
It appears that I am unable to write to a characteristic that is 64 bytes long.

Any attempts to write to the characteristic seems to crash the board. I know that with BLE it's required for the communication to chunk the writing. My app works with another BLE chip set but doesn't work with the Curie any hints?

BLECharacteristic credentials("bfe433cf-6b5e-4368-abab-b0a59666a403", BLERead | BLEWrite, 64);

As you can see this characteristic is 64 bytes long. The firmware works so long as no attempt is made to write to it.

sgbihu:
Which API you used? I test with .writeValue, it doesn't crash. So can you share your test sketches?

SidLeung:
@sgbihu , please share your sketch here for comparison.  Thanks.

kgillespieanaren:
This would be a write coming from the central not internal on the embedded device it's self. Yes .writeValue works, write with response from a central device to the Curie over BLE does not.

I will send my example but there isn't much to it.

sgbihu:
My test key code is:

`unsigned char test_1[64];
credentials.writeValue(test_1, sizeof(test_1));`

By the way, the stack only send out the 20 Bytes. Because the BLE protocol has this limitation. You'd better not use 64 Bytes.

kgillespieanaren:
So the Curie BLE stack doesn't handle long writes? There are methods in many BLE stacks, Android and iOS both support, to send what is know as a long write, sometime queued writes, your data is automatically chunk-ed into 20 bytes segments.

As far as I know the central sends a begin reliable write message, and then call write characteristics for as many 20 bytes segments as it needs then the peripheral device will appended that data to the characteristic until the central then sends the end reliable write message.

This feature exists and works well with other chip sets. Nordic has an example here https://github.com/NordicSemiconductor/ble_app_hrs---LongWrite

But I am guessing I can't expect this feature anytime soon for Curie?

sandeepmistry:
Hi @kgillespieanaren,

It would be great if you can share a minimal sketch that can be used by us to reproduce the issue. As well as steps that we can use to reproduce the issue.

@SidLeung @sgbihu Last time I checked, iOS tries to negotiate an MTU of 158 bytes, need to see if the CurieBLE accepts this but it might fallback to the minimum MTU of 23 bytes, which will trigger the long or reliable write on the iOS or Android central side.

[...]

PR #462, opened 2017-03-02
GIT #456 large characteristics crash board when attempting to write t…

sgbihu:
…o them

Root cause:
1. The typecast order is not correct make the class's offset has an error
    and make the memcpy used wrong address and wrong length.

Solution:
1. Adjust the typecast order, from base type to child type, in BLECallbacks.cpp


SidLeung:
This PR is replaced by #463 and closing this now.

PR #463, opened 2017-03-02
Jira 868, BLE Peripheral stack crash with long write, git456

SidLeung:
Root cause:
  - Upon processing the write event, The received user data
    was casted incorrectly using the base class.  Resulted
    in using an incorrect offset that caused memcpy to
    use incorrect address and length.

Code Mods:
  1. BLECallbacks.cpp:
     Adjust the typecast order, from base class to child
     class in profile_longwrite_process() and
     profile_longflush_process().

### DIFF
--- a/libraries/CurieBLE/src/internal/BLECallbacks.cpp
+++ b/libraries/CurieBLE/src/internal/BLECallbacks.cpp
@@ -77,7 +77,13 @@ ssize_t profile_longwrite_process(struct bt_conn *conn,
                                      const void *buf, uint16_t len,
                                      uint16_t offset)
 {
-    BLECharacteristicImp *blecharacteritic = (BLECharacteristicImp*)attr->user_data;
+    BLEAttribute *bleattr = (BLEAttribute *)attr->user_data;
+    BLEAttributeType type = bleattr->type();
+    if (BLETypeCharacteristic != type)
+    {
+        return 0;
+    }
+    BLECharacteristicImp *blecharacteritic = (BLECharacteristicImp*)bleattr;
     
     blecharacteritic->setBuffer((const uint8_t *) buf, len, offset);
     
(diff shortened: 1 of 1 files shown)


SidLeung:
This PR addresses the Peripheral long characteristic write issue.

@russmcinnis @noelpaz , please perform system testing.  You can verify the fix with the sample sketch in the gitHub issue #456.

libraries/CurieBLE/src/internal/BLECallbacks.cpp
```diff
@@ -77,7 +77,13 @@ ssize_t profile_longwrite_process(struct bt_conn *conn,
                                      const void *buf, uint16_t len,
                                      uint16_t offset)
 {
-    BLECharacteristicImp *blecharacteritic = (BLECharacteristicImp*)attr->user_data;
+    BLEAttribute *bleattr = (BLEAttribute *)attr->user_data;
```
eriknyquist: if it is possible for the user to set this as NULL, then I think we should check the pointer before de-referencing, to avoid a crash.

libraries/CurieBLE/src/internal/BLECallbacks.cpp
```diff
@@ -88,7 +94,13 @@ int profile_longflush_process(struct bt_conn *conn,
                               const struct bt_gatt_attr *attr, 
                               uint8_t flags)
 {
-    BLECharacteristicImp *blecharacteritic = (BLECharacteristicImp*)attr->user_data;
+    BLEAttribute *bleattr = (BLEAttribute *)attr->user_data;
+    BLEAttributeType type = bleattr->type();
```
eriknyquist: Same as line 81

libraries/CurieBLE/src/internal/BLECallbacks.cpp
sgbihu: I think we can use assert. Because this struct not expose to user. The CurieBLE library own this memory. How about this?

SidLeung: @sgbihu , we are looking for the fact that whether, attr->user_data, could be a NULL or not.  And from your answer, it indicates that content of user_data is a pointer to a memory that is pre-allocated and, thus, would not be NULL.  I think the label, user_data, has a different meaning here and that alerts us.  Here, user_data, is actually a field called user_data in the BLE Characteristic.  Usually, in code, label such as, user_data, represents input from user (or application).  It is not the case here.

@sgbihu , do you agree with my explanation?

I don't think there is a need to change anything here.  @eriknyquist , would you agree?

eriknyquist: OK, sounds good to me

SidLeung:
Code change approved, testing completed, proceed with merging to main trunk.

@yashaswini-hanji , please merge this PR to main trunk and close out the Jira.

yashaswini-hanji:
PR merged and Jira closed

PR #465, opened 2017-03-04
Jira 870, Change sensortag_button.ino to use localName

SidLeung:
Feature added:
  - Since the BLE library now support the Scan
    Response Data processing, update the sketch
    to look for the sensorTag localName instead
    of its specific unit address.

Cod mods:
1. sensortag_button.ino:
   - Replace MAC address comparison with localName
     string comparison.


SidLeung:
@russmcinnis , change the senortag_button.ino as per your suggestion.  Did update the comment block to make sure the user is aware of the localName may vary.  Please test it.

SidLeung:
@yashaswini-hanji , please merge the PR to main trunk and close out the ticket.

PR #468, opened 2017-03-09
Modify comments to point to Paul Stoffregen's Serial Flash library

noelpaz:
The current comments in the CurieSerialFlash examples has the library manager shortcut and the user will be downloading the versions 1.0 to 4.0, but these versions will not work with these examples and compilation fails. The user instead should be instructed to get the current working library version from github at https://github.com/PaulStoffregen/SerialFlash


noelpaz:
@eriknyquist Please review. I am a bit confused at the white space additions. It could be my editor.

noelpaz:
@eriknyquist . My editor cleaned some of the whitespace it looks like. So I cleaned that up.

PR #469, opened 2017-03-10
[TEMP] remove test direcyory from CurieBLE examples

yashaswini-hanji:


yashaswini-hanji:
@eriknyquist Please review it

Issue #368, opened 2016-12-29
CurieBLE: BLE.scan() should scan with duplicate filter


sandeepmistry:
In https://github.com/01org/corelibs-arduino101/blob/master/libraries/CurieBLE/examples/central/scan/scan.ino

```arduino
BLE.scan();
```

scans without filter, however I expect it to scan with a duplicate filter.

SidLeung:
Track by Jira 802

kitsunami:
Addressed by #457

sandeepmistry:
This is a bit better with the 2.0.1 RC1 release, however the logic to enable the duplicate filtering is inverted.

I've submitted PR #476 to address this.

SidLeung:
Please note that PR #476 is superseded by the follow PR,

https://jira.ndg.intel.com/browse/ATLEDGE-896

sandeepmistry:
This looks good now with the 2.0.2 RC JSON @SidLeung provided me. Closing

The scan.ino example only reports discovered peripherals once as expected and the scan_callback.ino example reports every advertisement.

However, as @SidLeung mentioned `BLE.scan()` does not report the scan response as I expected. I'll open a separate issue to discuss this.

PR #476, opened 2017-03-15
Correct duplicate filter inversion

sandeepmistry:
Scanning with duplicates should disable the filter, scanning without duplicates should enable the filter.

Related to #368.

### DIFF
--- a/libraries/CurieBLE/src/internal/BLEDeviceManager.cpp
+++ b/libraries/CurieBLE/src/internal/BLEDeviceManager.cpp
@@ -578,7 +578,7 @@ BLEDevice BLEDeviceManager::peripheral()
 
 bool BLEDeviceManager::startScanning()
 {
-    _adv_duplicate_filter_enabled = false;
+    _adv_duplicate_filter_enabled = true;
     _scan_param.filter_dup   = BT_HCI_LE_SCAN_FILTER_DUP_ENABLE;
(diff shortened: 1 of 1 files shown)


libraries/CurieBLE/src/internal/BLEDeviceManager.cpp
```diff
@@ -578,7 +578,7 @@ BLEDevice BLEDeviceManager::peripheral()
 
 bool BLEDeviceManager::startScanning()
 {
-    _adv_duplicate_filter_enabled = false;
```
sgbihu: I don't think you can change code here. If you changed code here, it will make developer confusing. I think change the code in BLEDevice.cpp is more reasonable.

russmcinnis:
I was actually expecting scan(true) to turn on the duplicate filter so you see devices only once. Is the following the correct behavior? 
@sandeepmistry @SidLeung 
This is from testing the json that merges PR 476, 478 and 480 which merges a few PRs.
BLE.scan(true); - no filter so same devices are re-scanned
BLE.scan(false); - filter duplicates so you only see devices once
BLE.scan() - filter duplicates so you only see devices once

russmcinnis:
system test passed

SidLeung:
This PR is replaced by the following PR,

https://github.com/01org/corelibs-arduino101/pull/491

PR #483, opened 2017-03-16
Make sure LED is not lit on sketch start

bigdinotech:
-The LED retains the last state it was in after a sketch upload
-We need to make sure its is not lit on sketch startup so that behavior
is consistent

### DIFF
--- a/variants/arduino_101/variant.cpp
+++ b/variants/arduino_101/variant.cpp
@@ -180,6 +180,8 @@ void variantGpioInit(void)
         SET_PIN_MODE(p->ulSocPin, p->ulPinMode);
         pinmuxMode[pin] = GPIO_MUX_MODE;
     }
+    //make sure the led is not lit on sketch start
+    digitalWrite(LED_BUILTIN, LOW);
 }
 
 void variantPwmInit(void)


bigdinotech:

SidLeung:
@russmcinnis , @noelpaz , please perform system testing.

bigdinotech:
https://github.com/01org/corelibs-arduino101/commit/2daed478047ff734551afcaddffb27debc352be3

sandeepmistry:
@bigdinotech nice!

Are all the pins in input mode on sketch start? That's the default behaviour of AVR boards like the Uno and we should aim for the same for the 101.

For example, the SAMD core for boards like the Zero and MKR1000 has the following: 

```c
  // Setup all pins (digital and analog) in INPUT mode (default is nothing)
  for (uint32_t ul = 0 ; ul < NUM_DIGITAL_PINS ; ul++ )
  {
    pinMode( ul, INPUT ) ;
  }
```

https://github.com/arduino/ArduinoCore-samd/blob/96745cd48d8232ed37ec03d3a71d510aa9410f38/cores/arduino/wiring.c#L83-L87

cc/ @cmaglie

PR #484, opened 2017-03-17
Jira 890, Allow BLE Descriptor reading to be blocking, git 478

### LINKED ISSUE #478: CurieBLE: descriptor.read() does not block
Similar to #383 but for descriptor.

I would like to remove the `delay(1000);` statements from the peripheral_explorer example. So, expect ` descriptor.read()` to behave like `characteristic.read()` and block.

sgbihu:
Feature requested:
  To make the reading of the Descriptor of a BLE Characteristic a blocking read method.  The mechanism is similar to the the Characteristic read method.  However, for Descriptor read(), it is always a blocking read.

Code Mods:
1.  libraries/CurieBLE/src/internal/BLEDescriptorImp.cpp:
     -  Added a flag for the Descriptor read completion event to set in foreground.
     -  Check on the Descriptor read completion event in the background as an exit condition for the read method.

### DIFF
--- a/libraries/CurieBLE/src/internal/BLEDescriptorImp.cpp
+++ b/libraries/CurieBLE/src/internal/BLEDescriptorImp.cpp
@@ -172,6 +172,7 @@ bool BLEDescriptorImp::read()
 {
     int retval = 0;
     bt_conn_t* conn = NULL;
+    bool ret_bool = true;
     
     if (true == BLEUtils::isLocalBLE(_bledev))
     {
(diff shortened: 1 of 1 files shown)


SidLeung:

@russmcinnis @noelpaz , please perform system testing.

This feature is requested for Deneb.RC2 release.

russmcinnis:
System test passed PR490(471, 477, 480, 484 and 489). will unit test next.

SidLeung:
@russmcinnis , please save your traces in the Jira ticket for record.

PR #472, opened 2017-03-15
Jira 798, Adjust ARC SRAM to BLE scan the SensorTag, git 373

sgbihu:
Feature requested:
    Arduino requires the BLE stack, in Central mode, be able to completely record all Attributes from the TI SensorTag.  There about 100+ Attributes and each requires buffering space in additional to the normal memory consumption of storing an Attribute.  Thus, a substantial increase in the HEAP memory (SRAM) size is needed.
   Please note that the change in SRAM arrangement affects both cores:  x86 and ARC.  Thus, image for both cores need to be updated together.  This PR addresses only the ARC core.

Code Mods:
  1.  flash.ld - Update the RAM memory map to reflect the new ARC SRAM starting memory.

WARNING:
  -  This PR needs the x86 f/w to be updated with this PR,
        https://github.com/01org/CODK-M-X86/pull/24

PR #494, opened 2017-03-23
Change SRAM mapping and increase SRAM allocated to ARC core

bigdinotech:
-increase SRAM to 40K
-change SRAM start address
-increase heap size to 16K
-this change is needed for BLE scan for SensorTag
-requires firmware update on x86 side

### DIFF
--- a/variants/arduino_101/linker_scripts/flash.ld
+++ b/variants/arduino_101/linker_scripts/flash.ld
@@ -40,7 +40,7 @@ OUTPUT_FORMAT("elf32-littlearc", "elf32-bigarc", "elf32-littlearc")
 MEMORY
     {
     FLASH                 (rx) : ORIGIN = 0x40034000, LENGTH = 152K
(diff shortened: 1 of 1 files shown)


bigdinotech:
@sgbihu after some discussions with @SidLeung, we decided to allocate 40K to ARC
@SidLeung @eriknyquist please review

Please do not merge until corresponding change in firmware is ready and firmware update tool is ready

bigdinotech:
replace #472

variants/arduino_101/linker_scripts/flash.ld
```diff
@@ -52,7 +52,7 @@ __firq_stack_size = 1024;
 
 /* Minimum heap size to allocate
  * Actual heap size might be bigger due to page size alignment */
-__HEAP_SIZE_MIN = 8192;
+__HEAP_SIZE_MIN = 16384;
```
sgbihu: I think we can add another 8K RAM for heap. Because ~8K RAM for user sketch may enough and now we have 40K RAM

PR #496, opened 2017-03-24
Fix Jira 864 After BLE.end() a peripheral is still showing up in scans

sgbihu:
Also fixed Jira 839
1. Stop scan, advertising and disconnect the connected devices when call end

Changed files
BLEDeviceManager.cpp - do the real action
BLEDevice.cpp - only local BLE device call the real end


russmcinnis:
This PR fixes the issue so now the peripheral stops advertising after BLE.end().

PR #500, opened 2017-03-24
Wait for scan response before reporting discovered peripheral

sandeepmistry:
This change waits for a scan response from the peripheral before reporting it to the user.

This is a proposal to fix https://github.com/01org/corelibs-arduino101/issues/497 and https://github.com/01org/corelibs-arduino101/issues/370#issuecomment-289112547.

The downside is that devices that don't respond to scan responses will be ignored.

@SidLeung please consider this for the Deneb release.

PR #502, opened 2017-03-27
Fix GIT 497 CurieBLE: SensorTag example no longer works out of the box

sgbihu:
Add mechanism to promise the scan response can be returned.

Changed Files
  BLEDeviceManager.cpp - Change the API to return the device when scan response changed
  BLEDeviceManager.h - Modify the API

The issues need to be discussed:
The SensorTag will return twice. The ADV and Scan response.

### DIFF
--- a/libraries/CurieBLE/src/internal/BLEDeviceManager.cpp
+++ b/libraries/CurieBLE/src/internal/BLEDeviceManager.cpp
@@ -1370,7 +1370,9 @@ BLEDevice BLEDeviceManager::available()
     {
         uint64_t timestamp_delta = timestamp - _peer_adv_mill[i];
         temp = &_peer_adv_buffer[i];
-        if ((timestamp_delta <= 2000) && (max_delta < timestamp_delta))
+        if ((timestamp_delta >= 800) && // Wait scan response coming  || _peer_scan_rsp_data_len[i] >= 0
+            (timestamp_delta <= 2000) && // Check timeout
+            (max_delta < timestamp_delta))
         {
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@SidLeung I prefer the behaviour of #500 over this.

I get 2 events from the SensorTag, testing with the scan example:

```
Discovered a peripheral
-----------------------
Address: 68:C9:0B:06:BC:81
Service UUID's: aa80 
RSSI: -62

Discovered a peripheral
-----------------------
Address: 68:C9:0B:06:BC:81
Local Name: CC2650 SensorTag
Service UUID's: aa80 
RSSI: -62
```

however with #500 I just get a single:

```
Discovered a peripheral
-----------------------
Address: 68:C9:0B:06:BC:81
Local Name: CC2650 SensorTag
Service UUID's: aa80 
RSSI: -62
```

which is what I expect.

sgbihu:
Yes. That's why I add the comment at the end of this PR. I think both your PR and mine are not perfect solution. 

This is behavior's conflicts. The peripheral has two part in advertise step, the advertising and scan response are come in different time stamp. And some connectable peripheral also can not send out the scan response. We can't define a policy to adapt all scenario when enable the duplicate filter.

I suddenly have another idea to use time stamp to return the device when call available. The scan request will send out after the central see the advertising. So the scan response will be got in an expected time. The implementation can add a delay to return new device. I will try to implement a solution and you can test again.

SidLeung:
This PR will continue with #508.

PR #510, opened 2017-03-29
Jira 714 Find the limits of the number of Services and Characteristics

sgbihu:
Make the BLE support more characteristics and services by separate the
  service registration

Changed files
  BLEProfileManager.cpp - Remove register all service at same time
  BLEProfileManager.h - Change the comments
  BLEServiceImp.cpp - Add the register service after update the characteristic


SidLeung:
@bigdinotech and @eriknyquist.  Please review the code change.

PR #513, opened 2017-03-29
Renamed examples

agdl:
In arduino style examples name are without _ and with camel case first capital letter


sandeepmistry:
@SidLeung a small change to be considered for Deneb

PR #514, opened 2017-03-29
Added missing keyword

agdl:
In this way CurieBLE is highlighted as all the Arduino .h are


sandeepmistry:
@SidLeung another minor change to consider for Deneb. Thanks.

Issue #516, opened 2017-03-29
CurieBLE: characteristic.read() reports success when peripheral responds with error


sandeepmistry:
Testing with the 2.0.1RC2.1 JSON:

Arduino sketch:

```arduino
#include <CurieBLE.h>

void setup() {
  Serial.begin(9600);

  while (!Serial);

  Serial.println("BLE Central - Read + Write test");

  // initialize the BLE hardware
  BLE.begin();

  // start scanning for peripheral
  BLE.scanForName("rwfail");
}

void loop() {
 // check if a peripheral has been discovered
  BLEDevice peripheral = BLE.available();

  if (peripheral) {
    // discovered a peripheral, print out address, local name, and advertised service
    Serial.print("Found ");
    Serial.println(peripheral.address());

    BLE.stopScan();

    Serial.println("Connecting ...");
    if (peripheral.connect()) {
      Serial.println("Connected");
    } else {
      Serial.println("Failed to connect!");
      while(1);
    }
    
    // discover peripheral attributes
    Serial.println("Discovering attributes ...");
    if (peripheral.discoverAttributes()) {
      Serial.println("Attributes discovered");
    } else {
      Serial.println("Attribute discovery failed.");
      peripheral.disconnect();
      while(1);
    }

    BLECharacteristic c = peripheral.characteristic("fa12");


    while (peripheral.connected()) {
      Serial.println("reading ...");
      if (c.read()) {
        // should not happen
        Serial.println("read success");
      } else {
        // expected
        Serial.println("read fail");
      }

      Serial.println("writing ...");
      if (c.writeInt(42)) {
        // should not happen
        Serial.println("write success");
      } else {
        // expected
        Serial.println("write fail");
      }


      delay(1000);
    }


    BLE.scan();
  }
}
```

Node.js bleno test app:

```js
var bleno = require('bleno');

var BlenoPrimaryService = bleno.PrimaryService;
var BlenoCharacteristic = bleno.Characteristic;

console.log('bleno - read + write fail');

bleno.on('stateChange', function(state) {
  console.log('on -> stateChange: ' + state);

  if (state === 'poweredOn') {
    bleno.startAdvertising('rwfail', ['fa11']);
  } else {
    bleno.stopAdvertising();
  }
});

bleno.on('advertisingStart', function(error) {
  console.log('on -> advertisingStart: ' + (error ? 'error ' + error : 'success'));

  if (!error) {
    bleno.setServices([
      new BlenoPrimaryService({
[...]

PR #518, opened 2017-03-31
Fix GIT #516 and 517 return success when respond with error

sgbihu:
Root cause
1. The return value indecate the read/write request send successfully.
2. The subscribe will ignore the response and this will not be fixed.

Solution
1. Add the error detect based on read/write response.

Changed files
BLECallbacks.cpp - Delete the error check for read response.
BLECharacteristicImp.cpp
  - Add the error check.
  - Optimize the memory allocate and clear the allocated memory.
BLECharacteristicImp.h - Add varibles to flag the response error

### DIFF
--- a/libraries/CurieBLE/src/internal/BLECallbacks.cpp
+++ b/libraries/CurieBLE/src/internal/BLECallbacks.cpp
@@ -157,10 +157,6 @@ uint8_t profile_read_rsp_process(bt_conn_t *conn,
                                  const void *data, 
                                  uint16_t length)
 {
-    if (NULL == data && 0 != length)
-    {
-        return BT_GATT_ITER_STOP;
-    }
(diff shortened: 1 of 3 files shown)


sandeepmistry:
> The subscribe will ignore the response and this will not be fixed.

Will this be fixed later?

sgbihu:
This always ignore the response in the stack. So the application can't get the result. This is the stack's limitation. And the return value stand for the gatt ccc write request send out. And it is an unblock call. I think we will not fix for the subscribe API.

sandeepmistry:
Is this a limitation of the firmware?

> And it is an unblock call. I think we will not fix for the subscribe API.

I expect `characteristic.subscribe()` and `characteristic.unsubscribe()` to block.

PR #521, opened 2017-04-03
Fix compile rules for paths with spaces

facchinm:
The Java IDE applies specific transformations to the input "recipes" based on locale and OS. On Windows, if a path contains a space, the quotation marks get ignored and the whole commandline gets passed to the underlying GO program as a single string, leading to a failure when calling `flag.Parse` . Enclosing the whole flag+path instead than the path alone fixes the issue.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -71,7 +71,7 @@ recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compil
 recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} "-T{build.variant.path}/{build.ldscript}" "-Wl,-Map,{build.path}/{build.project_name}.map" {compiler.c.elf.extra_flags} -o "{build.path}/{build.project_name}.elf" "-L{build.path}" "-L{build.variant.path}" -Wl,--whole-archive "-l{build.variant_system_lib}" -Wl,--no-whole-archive -Wl,--start-group "-l{build.variant_system_lib}" -lnsim -lc -lm -lgcc {object_files} "{build.path}/{archive_file}"
 
 ## Save output with debug symbols (.debug.elf file). Uncomment if you wish to use OpenOCD to debug.
-recipe.hooks.objcopy.preobjcopy.1.pattern={runtime.tools.arduino101load.path}/arduino101load -c -from="{build.path}/{build.project_name}.elf" -to="{build.path}/../arduino101_sketch.debug.elf"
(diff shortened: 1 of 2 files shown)


SidLeung:
Will not use this PR in its entirety.  However, will cherry-pick your change for the compile rules into the main trunk code and the Deneb.RC2 tag manually.

platform.txt
```diff
@@ -71,7 +71,7 @@ recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compil
 recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} "-T{build.variant.path}/{build.ldscript}" "-Wl,-Map,{build.path}/{build.project_name}.map" {compiler.c.elf.extra_flags} -o "{build.path}/{build.project_name}.elf" "-L{build.path}" "-L{build.variant.path}" -Wl,--whole-archive "-l{build.variant_system_lib}" -Wl,--no-whole-archive -Wl,--start-group "-l{build.variant_system_lib}" -lnsim -lc -lm -lgcc {object_files} "{build.path}/{archive_file}"
 
 ## Save output with debug symbols (.debug.elf file). Uncomment if you wish to use OpenOCD to debug.
-recipe.hooks.objcopy.preobjcopy.1.pattern={runtime.tools.arduino101load.path}/arduino101load -c -from="{build.path}/{build.project_name}.elf" -to="{build.path}/../arduino101_sketch.debug.elf"
+recipe.hooks.objcopy.preobjcopy.1.pattern="{runtime.tools.arduino101load.path}/arduino101load" -c -from="{build.path}/{build.project_name}.elf" -to="{build.path}/../arduino101_sketch.debug.elf"
```
SidLeung: @facchinm ,
Would like to know if the, ", for the, -from=, flag should be moved to the front of the flag like the ones that are done in line 110 and 118?  I don't think it is going to make a different.  Thus, we are going with this PR for the Deneb release.  Just being curious.

facchinm: Hi Sid, the quotation here are "correct" (in the sense you can execute them in a shell and they get expanded correctly) because the `recipe.*` rules get "spawned" inside the (golang) builder, which doesn't do any strange mangling (as opposed to the `upload.*` rules which are executed by the Java IDE and exposed the bug). In fact, one of the next step we wanted to take is also rewriting the uploader in GO to get rid of this totally inexplicable difference :slightly_smiling_face:

SidLeung:
Changes were cherry-picked and merged to main trunk (#522).  Thus, closing this PR.

PR #525, opened 2017-04-04
Enable firmware version checking

bigdinotech:
-enables checking of firmware version before sketch upload


bigdinotech:
@yashaswini-hanji @russmcinnis To test, for now we would have to manually create a 2.0.4 directory inside arduiono101load then replace the quark.bin with the binary built from CODK-M

SidLeung:
Looks good, @bigdinotech .  @yashaswini-hanji , @russmcinnis , let generate a JSON with this PR to test it out.

Issue #507, opened 2017-03-28
BLE.connected() always return false.


jimaobian:
Here is the code. It always return false, even BLE is connected.

corelib Version: 2.0.1-RC2

```
/*
 * Copyright (c) 2016 Intel Corporation.  All rights reserved.
 * See the bottom of this file for the license terms.
 */

#include <CurieBLE.h>

const int ledPin = 13; // set ledPin to on-board LED
const int buttonPin = 4; // set buttonPin to digital pin 4

BLEService ledService("19B10010-E8F2-537E-4F6C-D104768A1214"); // create service


// create switch characteristic and allow remote device to read and write
BLECharCharacteristic ledCharacteristic("19B10011-E8F2-537E-4F6C-D104768A1214", BLERead | BLEWrite);
// create button characteristic and allow remote device to get notifications
BLECharCharacteristic buttonCharacteristic("19B10012-E8F2-537E-4F6C-D104768A1214", BLERead | BLENotify); // allows remote device to get notifications

void setup() {
  Serial.begin(9600);
  pinMode(ledPin, OUTPUT); // use the LED on pin 13 as an output
  pinMode(buttonPin, INPUT); // use button pin 4 as an input

  // begin initialization
  BLE.begin();

  // set the local name peripheral advertises
  BLE.setLocalName("BtnLED");
  // set the UUID for the service this peripheral advertises:
  BLE.setAdvertisedService(ledService);

  // add the characteristics to the service
  ledService.addCharacteristic(ledCharacteristic);
  ledService.addCharacteristic(buttonCharacteristic);

  // add the service
  BLE.addService(ledService);

  ledCharacteristic.setValue(0);
  buttonCharacteristic.setValue(0);

  // start advertising
  BLE.advertise();

  Serial.println("Bluetooth device active, waiting for connections...");
}

void loop() {

//////////////////////////
  static unsigned long BLEWriteTimer = millis();
  if (millis() - BLEWriteTimer >= 1000) {
    BLEWriteTimer = millis();
    Serial.println(BLE.connected());
  }
//////////////////////////
  
  // poll for BLE events
  BLE.poll();

  // read the current button pin state
  char buttonValue = digitalRead(buttonPin);

  // has the value changed since the last read
  boolean buttonChanged = (buttonCharacteristic.value() != buttonValue);

  if (buttonChanged) {
    // button state changed, update characteristics
    ledCharacteristic.setValue(buttonValue);
[...]

PR #454, opened 2017-02-23
Jira 857, BLEPeripheral::connected() always returns false, git 444

SidLeung:
Root casue:
  - The stack need use connected device to get the state.
  - The current code used local address to get the link and get an error.

Code mods:

1. BLEPeripheral.cpp:
   - Use central's address to get connection state.

PR #528, opened 2017-04-07
Correct value of BLE.connected()

sandeepmistry:
As discussed in #507.

Now it will check:
- if in peripheral mode and if a central is connected
- if in central mode and if a peripheral is connected

I've also reverted https://github.com/01org/corelibs-arduino101/commit/a2df1795ac275ac48485329dc0697ab2324715bf from PR #454 as the change it no longer required with https://github.com/01org/corelibs-arduino101/commit/58c323e278a510e6e6ec2d0c4f29de69a0c11384.

cc/ @jimaobian


SidLeung:
@sgbihu please review the code change as a solution to,

https://github.com/01org/corelibs-arduino101/issues/507#issuecomment-292660771

PR #533, opened 2017-04-20
Increase delay between each Serial.write()

bigdinotech:


bigdinotech:
@eriknyquist @SidLeung please review

PR #534, opened 2017-04-21
V4.2 Porting

sgbihu:
Porting V4.2 to improve the performance. Please don't merge.


libraries/CurieBLE/src/internal/BLEDeviceManager.h
```diff
@@ -392,7 +400,7 @@ class BLEDeviceManager
     uint8_t    _peer_adv_data[BLE_MAX_ADV_BUFFER_CFG][BLE_MAX_ADV_SIZE];
     uint8_t    _peer_adv_data_len[BLE_MAX_ADV_BUFFER_CFG];
     uint8_t    _peer_scan_rsp_data[BLE_MAX_ADV_BUFFER_CFG][BLE_MAX_ADV_SIZE];
-    int8_t     _peer_scan_rsp_data_len[BLE_MAX_ADV_BUFFER_CFG];
+    uint8_t    _peer_scan_rsp_data_len[BLE_MAX_ADV_BUFFER_CFG];
```
sandeepmistry: @SidLeung @sgbihu I think PR https://github.com/01org/corelibs-arduino101/pull/500 was reverted here. Is this change intended?

sgbihu: This change has been considered. You implementation has an assumption, the scan response should followed by advertisement. But this may has some corner case to make ADV not be reported to sketch.
Now I use timeout mechanism. The central will report ADV data after 800ms delay.

PR #538, opened 2017-05-03
pulseIn() pullup/down comment 

russmcinnis:
There is a situation where using pulseIn() on a pin that has a floating voltage could cause a sketch to hang at the pulseIn() function call. A comment was added to the pulseIn() function header to use a pull up or pull down resistor.


russmcinnis:
@SidLeung please review.

eriknyquist (approved):
Looks good to me, @SidLeung to review also

PR #541, opened 2017-05-10
Revert fixed cdc-acm delays

bigdinotech:
-fixed delay no longer needed by latest firmware


bigdinotech:
replaced by https://github.com/01org/corelibs-arduino101/pull/546

PR #544, opened 2017-05-11
CurieBLE: Correct EIR length calculations

### LINKED ISSUE #543: CurieBLE: advertising doesn't work with longer local names
Observed by @CarlaMSedze and @rodriguezcarlos from the Arduino CTC team in Malmo.

```arduino
#include <CurieBLE.h>

void setup() {
  BLE.begin();

  BLE.setLocalName("Messenger");

  BLE.advertise();
}

void loop() {
}
```

We expect this sketch to work. As with 2.0.2 both the advertisement data and scan response are supported. I think we should be able to support local name values up to 20 characters ....

sandeepmistry:
Fixes #543.

Corrects two things:

1) `lengthOfScanRsp` was never updated
2) Space calculations where not taking EIR header length (2 bytes) into account

@SidLeung please consider this change for Elnath, if it is too late for Bootes.

cc/ @CarlaMSedze @rodriguezcarlos

### DIFF
--- a/libraries/CurieBLE/src/internal/BLEDeviceManager.cpp
+++ b/libraries/CurieBLE/src/internal/BLEDeviceManager.cpp
@@ -369,17 +369,16 @@ BLEDeviceManager::setAdvertiseData(uint8_t type, const uint8_t* data, uint8_t le
     
     for (uint8_t i = 0; i < _scan_rsp_data_idx; i++)
     {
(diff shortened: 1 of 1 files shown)


sgbihu (approved):

SidLeung:
@russmcinnis , please perform system testing and inform @yashaswini-hanji  upon successful completion.  Thanks.

PR #547, opened 2017-05-15
Add IPM (Inter-Processor-Mailbox) driver to cores/arduino

eriknyquist:
This allows the mailbox to be used for Serial (instead of polling shared
memory). The CurieMailbox library, in the codk-m-corelibs branch, will
not require any API changes, but will now just be a thin wrapper around
the methods in mailbox.cpp

@bigdinotech, look at mailbox.h to see the functions available to you.
 use channel 7 for serial (e.g. `mailbox_register(7, callback);`)

### DIFF
--- a/cores/arduino/mailbox.cpp
+++ b/cores/arduino/mailbox.cpp
@@ -1,0 +1,163 @@
+#include <string.h>
+#include "interrupt.h"
+#include "scss_registers.h"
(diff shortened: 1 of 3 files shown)


eriknyquist:
Need to test the CODK mailbox examples first (Serial + mailbox being used at once)

PR #552, opened 2017-05-18
Use IPM for cdc-acm

bigdinotech:
-use the IPM instead of shared memory for passing data between ARC
and x86 cores for cdc-acm.

### DIFF
--- a/cores/arduino/CDCSerialClass.cpp
+++ b/cores/arduino/CDCSerialClass.cpp
@@ -30,11 +30,34 @@
 #include "wiring_digital.h"
 #include "variant.h"
 
-#define CDCACM_FIXED_DELAY     120
+#define CDC_MAILBOX_TX_CHANNEL  7
+#define CDC_MAILBOX_RX_CHANNEL  6
 
 extern void CDCSerial_Handler(void);
 extern void serialEventRun1(void) __attribute__((weak));
 extern void serialEvent1(void) __attribute__((weak));
+
+static void cdc_mbox_isr(CurieMailboxMsg msg)
+{
+    char *rxbuffptr = (char*)msg.data;
+    int new_head;
+    for(int i = 0; i < MBOX_BYTES; i++)
+    {
+        if((uint8_t)(*(rxbuffptr+i)) != '\0')
+        {
+            new_head = (Serial._rx_buffer->head +1) % CDCACM_BUFFER_SIZE;
+            if(new_head != Serial._rx_buffer->tail)
+            {
+                Serial._rx_buffer->data[Serial._rx_buffer->head] = *(rxbuffptr+i);
+                Serial._rx_buffer->head = new_head;
+            }
+        }
+        else
+        {
+            break;
+        }
+    }
+}
 
 // Constructors ////////////////////////////////////////////////////////////////
 
@@ -47,9 +70,9 @@ CDCSerialClass::CDCSerialClass(uart_init_info *info)
 
 void CDCSerialClass::setSharedData(struct cdc_acm_shared_data *cdc_acm_shared_data)
 {
-    this->_shared_data = cdc_acm_shared_data;
-    this->_rx_buffer = cdc_acm_shared_data->rx_buffer;
-    this->_tx_buffer = cdc_acm_shared_data->tx_buffer;
+    _shared_data = cdc_acm_shared_data;
+    _rx_buffer = cdc_acm_shared_data->rx_buffer;
+    _tx_buffer = cdc_acm_shared_data->tx_buffer;
 }
 
 void CDCSerialClass::begin(const uint32_t dwBaudRate)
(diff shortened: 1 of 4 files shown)


bigdinotech:
@eriknyquist @SidLeung ready for review

cores/arduino/CDCSerialClass.cpp
```diff
@@ -73,7 +100,9 @@ void CDCSerialClass::init(const uint32_t dwBaudRate, const uint8_t modeReg)
      * Empty the Rx buffer but don't touch Tx buffer: it is drained by the
      * LMT one way or another */
     _rx_buffer->tail = _rx_buffer->head;
-
+    
+    mailbox_register(6, cdc_mbox_isr);
```
eriknyquist: Channel 6 is exposed by the CurieMailbox library. If you use this channel, the Mailbox library and Serial will not work simultaneously.

Can we use channel 7 for both RX and TX?

cores/arduino/CDCSerialClass.cpp
```diff
@@ -36,6 +36,33 @@ extern void CDCSerial_Handler(void);
...
+{
+    int new_head;
+    for(int i = 0; i < 4; i++)
+    {
+        uint32_t dword = msg.data[i];
+        for(int j = 0; j <4; j++)
+        {
+            if(((uint8_t)dword) != '\0')
+            {
+                new_head = (Serial._rx_buffer->head +1) % CDCACM_BUFFER_SIZE;
```
eriknyquist: Surely there's some kind of 'ringbuffer_add' method? we should not be re-implementing this logic for each write to the buffer.

cores/arduino/CDCSerialClass.cpp
```diff
@@ -36,6 +36,33 @@ extern void CDCSerial_Handler(void);
 extern void serialEventRun1(void) __attribute__((weak));
 extern void serialEvent1(void) __attribute__((weak));
 
+static void cdc_mbox_isr(CurieMailboxMsg msg)
```
eriknyquist: This function looks really complex for such a simple task. Can't you just do something like this;

```
unsigned char *ptr = (unsigned char*) msg.data;

/* Continue until 16 chars, or until NULL is seen */
for (unsigned i = 0; i < 16 && ptr[i]; ++i) {
    /* Assuming we have some kind of 'put' method for ring buffer... */
    ringbuffer_put(ptr[i]);
}
```

cores/arduino/CDCSerialClass.cpp
```diff
@@ -131,28 +158,16 @@ void CDCSerialClass::flush( void )
     }
 }
 
-size_t CDCSerialClass::write( const uint8_t uc_data )
+size_t CDCSerialClass::write(uint8_t uc_data )
 {
-    uint32_t retries = 2;
-
+    CurieMailboxMsg cdcacm_msg;
+
```
eriknyquist: trailing whitespace

cores/arduino/CDCSerialClass.cpp
```diff
@@ -131,28 +158,16 @@ void CDCSerialClass::flush( void )
...
-            _tx_buffer->head = i;
-
-	    // Just use a fixed delay to make it compatible with the CODK-M based firmware
-	    delayMicroseconds(CDCACM_FIXED_DELAY);
-            break;
-        }
-    } while (retries--);
-
+    cdcacm_msg.channel = 7;
+    cdcacm_msg.data[0] = (uint32_t)uc_data;
```
eriknyquist: Why the cast to uint32_t? not needed at all. remove please

cores/arduino/CDCSerialClass.cpp
```diff
@@ -131,28 +158,16 @@ void CDCSerialClass::flush( void )
...
-            _tx_buffer->data[_tx_buffer->head] = uc_data;
-            _tx_buffer->head = i;
-
-	    // Just use a fixed delay to make it compatible with the CODK-M based firmware
-	    delayMicroseconds(CDCACM_FIXED_DELAY);
-            break;
-        }
-    } while (retries--);
-
+    cdcacm_msg.channel = 7;
```
eriknyquist: 7. Give it a name. You know the drill, by now I hope.....

cores/arduino/CDCSerialClass.cpp
bigdinotech: From my initial tests using channel 7 for both was not stable. It freezes when we are sending and receiving on both cores at a high rate

eriknyquist: Throughput would be vastly improved if we fill all 16 bytes of the mailbox channel, instead of sending 1 message per character. Let's focus on that first, and then see if it's stable enough to use one channel for RX/TX

cores/arduino/CDCSerialClass.cpp
bigdinotech: there is none implemented for the cdc-acm ring buffer in shared memory

eriknyquist: let's implement one, then.

PR #553, opened 2017-05-19
Slim down sketch size 

facchinm:
Since 0304de781d54d0158c070d2d9cc750ab295c0712 the sketch size has exploded. Moving the BLE stack to ARC core is the main responsible.
This patch addresses two main fields:
- avoid including useless functions (WIP)
- avoid including ble symbols in sketches which don't use it

The first patch could be further optimized but since most sketches uses the CDC port it doesn't make much sense.

Sketch size comparison:
BareMinimum: 48744 -> 12540
(BLE central) Scan: 91692 -> 80440


variants/arduino_101/linker_scripts/flash.ld
```diff
@@ -158,6 +158,9 @@ SECTIONS
     {
         . = ALIGN(4);
         __start_heap = .;
+        end = __start_heap;
+        _end = end;
+        __end = end;
```
eriknyquist: I'm not clear on the purpose of these 3 new symbols? can you explain them?

facchinm: This is the first step to properly support newlib's `nano.specs`, which unfortunately aren't compiled in the current toolchain. The nano lib doesn't provide all the symbols and the linker would fail with something like 
```
sbrk.c:(.text._sbrk+0x2): undefined reference to `end'
```

eriknyquist:
@SidLeung  can you review please?

PR #600, opened 2020-12-03
move extra_flags into flags

### LINKED ISSUE #599: Should extra_flags be empty in platform.txt?
In my continuous integration pipeline, I was using the arduino-cli to compile an Arduino library example for the Intel:arc32:arduino_101 like:

```
arduino-cli compile --build-properties compiler.cpp.extra_flags='-DIN_CI' --fqbn "Intel:arc32:arduino_101" "examples/Continuous"
```

The `IN_CI` define is supposed to let me modify my code example if I'm compiling in the CI.

However, I get this error message:

```ShellSession
/home/ryantm/.arduino15/packages/Intel/hardware/arc32/2.0.4/cores/arduino/error.h:24:19: fatal error: os/os.h: No such file or directory
 #include "os/os.h"
```

If I add in the [cpp.extra_flags](https://github.com/arduino/ArduinoCore-arc32/blob/00f50b3f231cee5f5bf1c9395e278160ab9edab1/platform.txt#L40), it compiles fine:

```ShellSession
$ arduino-cli compile --build-properties 'compiler.cpp.extra_flags=-DIN_CI -D__CPU_ARC__ -DCLOCK_SPEED=32 -DCONFIG_SOC_GPIO_32 -DCONFIG_SOC_GPIO_AON -DINFRA_MULTI_CPU_SUPPORT -DCFW_MULTI_CPU_SUPPORT -DHAS_SHARED_MEM -I{build.system.path}/libarc32_arduino101/common -I{build.system.path}/libarc32_arduino101/drivers -I{build.system.path}/libarc32_arduino101/bootcode -I{build.system.path}/libarc32_arduino101/framework/include' --fqbn "Intel:arc32:arduino_101" "examples/Continuous"
Sketch uses 20708 bytes (13%) of program storage space. Maximum is 155648 bytes.
```

I also noticed that the [avr core](https://github.com/arduino/ArduinoCore-avr/blob/60f0d0b125e06dbf57b800192c80e5f60d681438/platform.txt#L42) does not populate these extra_flags.

I haven't found documentation on it, but I believe that the extra_flags should be empty so that the local build system can use them.

ryantm:
resolves #599

By leaving extra_flags empty, we leave a hook for the end-user to use
without having to duplicate a bunch of flags. This matches how the AVR
core leaves the extra_flags blank.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -14,12 +14,12 @@ compiler.prefix=arc-elf32
 
 compiler.path={runtime.tools.arc-elf32.path}/bin/
 compiler.c.cmd=arc-elf32-gcc
-compiler.c.flags=-c -std=gnu11 -mcpu=quarkse_em -mlittle-endian -g -Os -Wall -fno-reorder-functions -fno-asynchronous-unwind-tables -fno-omit-frame-pointer -fno-defer-pop -Wno-unused-but-set-variable -Wno-main -ffreestanding -fno-stack-protector -mno-sdata -ffunction-sections -fdata-sections -fsigned-char -MMD -D__ARDUINO_ARC__ -DCONFIG_BLUETOOTH_PERIPHERAL -DCONFIG_BLUETOOTH_CENTRAL -DCONFIG_BLUETOOTH_GATT_CLIENT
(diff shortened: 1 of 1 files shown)


ryantm:
I have no evidence that any human is looking at this, but if they do; I'll consider the CLA at that point.

